PR #56, opened 2024-12-27
Implement flask/charm/jewel weights

sida-wang:
Adds weights for flask mods, charm mods and jewel mods. 

Hardcoded arrays in the mods.lua file seems like it's bloating up the file. Looking for advice on what the correct syntax is to move it out into a different file. Tried but couldn't get it working.

Radius jewel mods implementation is not complete as each mod needs to be tied to either small or notables. This information doesn't seem to be available in the dat files and will likely need to be manually entered through trade results.


src/Export/Scripts/ScriptResources/JewelModWeights.lua
```diff
@@ -0,0 +1,177 @@
+local strTypes = {
+    -- prefixes
+    "AreaOfEffect",
+    "GlobalPhysicalDamageReductionRatingPercent",
+    "ArmourBreak",
+    "Attackdamage",
```
Nightblade, line 6: Can/should this be `AttackDamage`?

sida-wang: Yep, typo from my end. Thanks for picking up on it. I'll submit a fix after work if not resolved by then

sida-wang: Fixed in #578

PR #429, opened 2025-01-19
Fix herald interaction with Coming Calamity

### LINKED ISSUE #253: Herald of Ice is not calculated as an aura/buff
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### Check for support

Windows

In game when Herald Of Ice is active and the passive Coming Calamity in the passive skill tree is allocated, all cold damage gets a 50% multiplier.

The calculation for Ice Strike is shown in the image

Image

In path of building, the 50% multiplier is not taken into account (from what I understand, Herald of Ice is not added as a buff) and the damage is less.

Image

In game, if I de-activate Herald of Ice, the calculations are exactly like in path of building.

Image

Nothing specific to reproduce really.

### Character build code

```shell
https://maxroll.gg/poe2/pob/k672067b
```

### Screenshots

_No response_

xspirus:
This commit fixes the issue where `Herald of Ash` and `Herald of Ice` were not considered buffs when calculating damage increases with the passive point `Coming Calamity`.

Fixes #253 .

Herald of Ice and Herald of Ash were not considered buffs, thus Coming Calamity was not applied to cold/fire damage respectively (i.e. 50% increased cold/fire damage).

- Compare to in game numbers.

https://maxroll.gg/poe2/pob/k672067b

### Before screenshot:

image

### After screenshot:

image

**P.S. 1:** There is also a weird bug where sometimes in order for Heralds to be considered buffs, I have to remove it from the skill gems and add it again, but I don't understand why.

**P.S. 2:** I am not sure that this fix is the right approach, I did my best by understanding other fixes in the codebase, hope this helps. Thank you for this amazing tool!


LocalIdentity:
The gem lua files need to be exported from the game files with changes to the same name .txt file

LocalIdentity (approved):

xspirus:
@LocalIdentity thank you for the help!

PR #435, opened 2025-01-20
Add Emotion filtering to Anoint popup

Quxxy:

This lets users select which emotions they have available to restrict the list of anoints.

I'd appreciate if someone could check that I'm getting the emotion icon images the correct way (via `main:LoadTree`, then accessing via `tree.ddsMap` directly).

The images are displayed by modifying the `CheckBoxControl` class to have a `checkImage` field that overrides the display.  I took some liberties with the display, giving enabled controls a white outline to make it clearer when a control is selected/not selected without making it excessively difficult to see the image.

I slightly refactored the `NotableDBControl` class so that the filtering controls are included in the `rect` for the control, as opposed to technically appearing outside of it (which was weird and required offsets in the code that uses the control).

- Toggle each of the emotions on and off, ensuring that listed anoints only have the selected emotions in the recipe.

N/A

### Before screenshot:

image

### After screenshot:

image

### DIFF
--- a/src/Classes/NotableDBControl.lua
+++ b/src/Classes/NotableDBControl.lua
@@ -12,6 +12,8 @@ local m_floor = math.floor
 local m_huge = math.huge
 local s_format = string.format
 
+local emotionList = {"Ire", "Guilt", "Greed", "Paranoia", "Envy", "Disgust", "Despair", "Fear", "Isolation", "Suffering"}
+
 ---@param node table
 ---@return boolean
 local function IsAnointableNode(node)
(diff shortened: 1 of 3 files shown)


src/Classes/NotableDBControl.lua
```diff
@@ -32,15 +36,52 @@ local NotableDBClass = newClass("NotableDBControl", "ListControl", function(self
...
 	self.controls.searchMode = new("DropDownControl", {"LEFT",self.controls.search,"RIGHT"}, {2, 0, 100, 18}, { "Anywhere", "Names", "Modifiers" }, function(index, value)
 		self.listBuildFlag = true
 	end)
+
+	-- Create and set up emotion filters.
+	local function getEmotionImages()
+		local tree = main:LoadTree(latestTreeVersion)
+		local images = {}
+		for _, emotionName in ipairs(emotionList) do
+			images[emotionName] = tree.ddsMap[emotionName]
```
justjuangui, line 54: Please use tree:GetAssetByName(emotionName) instead.

this is the best way to go :)

justjuangui (changes requested):

LocalIdentity (approved):

PR #446, opened 2025-01-20
Added the ability to custom change max node depth for heat map

DoubtinGiyov:
At the moment, you can change the values ​​for heatmap calculation to {5, 10, 15, all}. In some cases, this is inconvenient. Sometimes you want to search up to 4 or 6 or 20, etc.

I added a new value "Custom". By selecting it, a new EditControl appears, in which you can enter any value.

### Before screenshot:
Screenshot_1

### After screenshot:
Screenshot_2

### DIFF
--- a/src/Classes/TreeTab.lua
+++ b/src/Classes/TreeTab.lua
@@ -24,6 +24,7 @@ local TreeTabClass = newClass("TreeTab", "ControlHost", function(self, build)
 
 	self.build = build
 	self.isComparing = false;
+	self.isCustomMaxDepth = false;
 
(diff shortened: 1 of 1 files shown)


LocalIdentity:
Check how this looks on the minimum screen width, cause I'm guessing its going to go off the screen

LocalIdentity:
Yeah we should make the tree version dropdown smaller and move the search box up to the first row after it
Also if you're not in it already, we have a PoB discord send me a message / friend request (LocalIdentity) if you want an inv

DoubtinGiyov:
I fixed visibility at minimum screen width
image

PR #432, opened 2025-01-20
Fix shock min-max limits

Rezzemy:
Fixes #426.

### Description of the problem being solved:

Shock min-max limits were changed in commit 0b90f10 causing any shock applied to be applied at a magnitude of 20%, this pull request reverts the changes to those limits.

EDIT 1/19:
> When manually configuring the Effect of Shock in the Configuration panel, that number is processed and subject to the maximum and minimum % amounts for the shock ailment, so 35% will be processed and turned into 20% for damage calculation, which isn't correct.

### Link to a build that showcases this PR:
https://maxroll.gg/poe2/pob/4h8t50yo

### Configuration screenshot
image

### Before screenshot:
image

### After screenshot:
image

PR #496, opened 2025-01-21
Increase Shock Maximum per Stack to 100%

Saeldur:
Increases Maximum Shock per stack to 100%, as you cannot set it above 20% currently in POB.

https://pobb.in/8gNhxMRJ7QWU

### Before screenshot:
before

### After screenshot:
after


QuickStick123:
You should probably just implement shock magnitude instead. But this may be a good stop gap measure. https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/432#issuecomment-2601212364

Saeldur:
> You should probably just implement shock magnitude instead. But this may be a good stop gap measure. [#432 (comment)](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/432#issuecomment-2601212364)

Yeah, ideally that gets done but this is at least a stop gap measure to allow people to manually override it until someone updates Ailment calculations.

LocalIdentity:
Superseded by https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/628

PR #509, opened 2025-01-21
DRAFT: Reduce performance issues with Time-Lost Jewels

Peechey:
Investigating solutions to fix the performance issues with Time-Lost Jewel logic

-
-
-


Wires77:
Guessing this can be closed in favor of #555

PR #545, opened 2025-01-22
Fix MoM order when EnergyShieldProtectsMana is disabled

AndreVallestero:
Fix MoM order when EnergyShieldProtectsMana is disabled

Previously, Energy Shield was taking damage before Mana, even when EnergyShieldProtectsMana was disabled. This change fixes that by doing the MoM mitigation first unless the flag is enabled.


LocalIdentity:
Damage is still taken from energy shield first as energy shield protects life

PR #554, opened 2025-01-23
Add support for misc mods in Uniques and Tree

Blitz54:
EDIT: Check both comments plz

The passive tree had multiple nodes for "%Increased/Reduced Projectile Speed for Spell Skills", at the top right of the tree, notables being "Turn the Clock Forward" and "Turn the Clock Back". While adding the Chance of 2 Add Projs, I realized the quality on Fireball also adds chance of 2 projs, so that seems to calculate too now.

The same cluster as well as another south west of it, had "% chance for Spell Skills to fire 2 additional Projectiles"

I figured a separate text box for the 2 additional projectile chance was a better option than tacking on fractions of a projectile to the projectile count. I decided on "2 Add. Proj. Chance" for the short description, although suggestions appreciated. I avoided using +2 because I feel that would cause confusion with +2 skills. The other names in the code like "TwoAdditionalProjectilesChance" feel long but necessary, as it's not just a chance of ONE, but two projs. So if GGG adds any mods that give a chance of one additional proj, it would have to be implemented separately.

-Tried to copy previous implementation of "Terrain Chain" to keep everything uniform.
-Added a spell flag to the % chance 2 proj, so that it doesn't apply to bow skills like snipe or LA.
-Tested ball lighting and ember fusilade in game to confirm it does apply to the skill. They work, and I'm assuming GGG has been consistent with this.

https://maxroll.gg/poe2/pob/3jkov05o
### Before screenshot:
image

### After screenshot:
image
image


Blitz54:
Ended up finding a bunch of other surface level stuff I could tackle.
- Blind chance shown in "Other effects" in calc.
- +1 grenade projectile from "Cluster Bombs"
- +2 chain from "Mist Whisper" unique crossbow
- cannot be chilled or frozen from "Dream Fragments"
- "Long Fuse" support 30% more damage 
- "increased effect of auras from your aura skills"
- "reduced effect of ignite on you"
- "duration of ignite, shock and chill on enemies"
- "damage penetrates 10% of lightning resistance if on low mana" Jacked the bosses resistance super high and it does affect the dps when I set low mana, is there anywhere that shows total elemental penetration?

Think that's all, or very close. Let me know if there's anything I could do better.

image
image
image
image
image
image
image
image

src/Modules/ModParser.lua
```diff
@@ -408,6 +408,7 @@ local modNameList = {
 	["knockback distance"] = "EnemyKnockbackDistance",
 	-- Auras/curses/buffs
 	["aura effect"] = "AuraEffect",
+	["effect of auras from your aura skills"] = "AuraEffect",
```
LocalIdentity, line 411: Needs a check for the Aura skilltype

src/Modules/ModParser.lua
```diff
@@ -759,6 +764,8 @@ local modNameList = {
 	["bleed duration"] = { "EnemyBleedDuration" },
 	["bleeding duration"] = { "EnemyBleedDuration" },
 	["bleed duration on you"] = "SelfBleedDuration",
+	["to blind enemies on hit"] = { "BlindChance" },
+	["Supported skills have 100% chance to blind enemies on hit"] = { "BlindChance" },
```
LocalIdentity, line 768: Guessing you put this here for the blind support gem, handling for that is done in the skill gem which we already handle

src/Modules/ModParser.lua
```diff
@@ -4330,6 +4343,7 @@ local specialModList = {
 	["cannot be stunned if you have at least (%d+) crab barriers"] = function(num) return { flag("StunImmune", { type = "StatThreshold", stat = "CrabBarriers", threshold = num }), } end,
 	["cannot be blinded"] = { flag("Condition:CannotBeBlinded"), flag("BlindImmune") },
 	["cannot be blinded while affected by precision"] = { flag("Condition:CannotBeBlinded", { type = "Condition", var = "AffectedByPrecision"}), flag("BlindImmune", { type = "Condition", var = "AffectedByPrecision"}) },
+	["cannot be blinded while on full life"] = { flag("Condition:CannotBeBlinded", { type = "Condition", var = "FullLife"}), flag("BlindImmune", { type = "Condition", var = "FullLife"}) },
```
LocalIdentity, line 4346: You are not blind immune on full life, just cannot be blinded. You could be blinded at 90% of life and then heal to full and still be blind

src/Modules/ModParser.lua
```diff
@@ -4113,6 +4125,7 @@ local specialModList = {
 	["projectiles deal (%d+)%% increased damage with hits and ailments for each enemy pierced"] = function(num) return { mod("Damage", "INC", num, nil, 0, bor(KeywordFlag.Hit, KeywordFlag.Ailment), { type = "PerStat", stat = "PiercedCount" }, { type = "SkillType", skillType = SkillType.Projectile }) } end,
 	["(%d+)%% increased bonuses gained from equipped quiver"] = function(num) return {mod("EffectOfBonusesFromQuiver", "INC", num)} end,
 	["(%d+)%% increased bonuses gained from equipped rings"] = function(num) return {mod("EffectOfBonusesFromRings", "INC", num)} end,
+	["(%d+)%% chance for spell skills to fire 2 additional projectiles"] = function(num) return { mod("TwoAdditionalProjectilesChance", "BASE", num, nil , ModFlag.Spell) } end,
```
LocalIdentity, line 4128: Would be good to integrate this into the main projectile count box instead of being separate as we use the projectile count of a skill in a number of areas for DPS calculations

LocalIdentity (approved):

src/Modules/ModParser.lua
LocalIdentity: Fixed in my commit

LocalIdentity:
I made a couple fixes to your PR before merging

src/Modules/ModParser.lua
```diff
@@ -4343,7 +4342,7 @@ local specialModList = {
 	["cannot be stunned if you have at least (%d+) crab barriers"] = function(num) return { flag("StunImmune", { type = "StatThreshold", stat = "CrabBarriers", threshold = num }), } end,
 	["cannot be blinded"] = { flag("Condition:CannotBeBlinded"), flag("BlindImmune") },
 	["cannot be blinded while affected by precision"] = { flag("Condition:CannotBeBlinded", { type = "Condition", var = "AffectedByPrecision"}), flag("BlindImmune", { type = "Condition", var = "AffectedByPrecision"}) },
-	["cannot be blinded while on full life"] = { flag("Condition:CannotBeBlinded", { type = "Condition", var = "FullLife"}), flag("BlindImmune", { type = "Condition", var = "FullLife"}) },
```
Blitz54: @LocalIdentity coming back to this line. BlindImmune is used for the CalcSections page for "Blind Avoidance Chance". And in CalcDefence.lua, although I'm not sure on how it's used there exactly. Any specific reason why there is 2 flags like that? Is it possible/worth it to combine into one flag?

Also, CalcPerform.lua uses CannotBeBlinded, but doesn't seem to prevent blind on line 505. 
https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/278ccf89e40def73ab25bc76876c59b28174afa5/src/Modules/CalcPerform.lua#L505
Toggling "are you blinded" is config lowers evasions by the 20%, or equipping and item with "you are blinded" also bypasses the check. I assume CannotBeBlinded should be priority right?
Adding `Condition:CannotBeBlinded` seems to make it function properly.

PR #557, opened 2025-01-23
changed "you've" to you have to enable

Blitz54:
Shadow Dancing recharge

Fixes # .

-
-
-


Peechey:
image
FYI, there's an example just below, that will catch you've and you have

Blitz54:
Aaaah that makes sense, closed this anyway cause I'm trying to screw around with forks and branches and whatnot. Working on solo projects is straight forward for me, but first time with VS Code and git so it's a learning experience. I assumed making a draft pull request and closing it instead of asking for review meant no one would see it, clearly not lol

PR #572, opened 2025-01-24
Update ailment hit calc to account for lucky dmg

pauloday:
Fixes # .

-
-
-


pauloday:
Yeah it seemed a little too easy when I was working on it, I was hoping someone just forgot to hook something up lol.

The average roll calculation is already similar to a lucky hit calculation. So you would just need some modification for the lucky case, but I'm not sure what it would be. I did some modeling and I'm pretty confident it'd just be some changes to the consts or maybe new coefficients in the current `ailmentRollAverage` calculation. I think the main problem is that by time the average roll calc is done all damage types have been combined in `calcAilmentSourceDamage`. But that fn doesn't have info about whether the hit is lucky either. So I think a solution would be something like:

1. Adding `output[damageType .. "LuckyChance"]` around line 3312
2. Passing the max and average stacks into `calcAilmentSourceDamage`
3. Calculating and returning the average roll from there

PR #577, opened 2025-01-24
Initial, curse aoe on tree and Windscream mod

Blitz54:
Fixes # .

Passive tree had a cluster of "increased area of effect of curses". They now properly apply to the radius in the Calc tab for curse skills.

Windscream Unique boots have the mod "curses have no activation delay". I tried to use OVERRIDE to set delay to 0, similar to resistances, but that didn't want to work. So instead I followed how Chaos Inoculation works, by using "MORE" -100. This appears to work fine, it properly overrides Whispers of Doom which doubles the delay.

Small issue is that with 0 delay, the stat is hidden in the calc tab. It does the same thing in game, but I think it would be clearer to have it show 0 in the calc tab, so the player can hover over it and see why it's instant. As well as see potentially wasted nodes. Regardless, thats above my skill set at the moment.

-Confirmed that the tree nodes in game do in fact apply to the curse AOE.


LocalIdentity (approved):
I fixed the display issue for you. You can have a look at my commit if you're interested

Blitz54:
> I fixed the display issue for you. You can have a look at my commit if you're interested

Thanks very much! Appreciates you.

PR #584, opened 2025-01-24
Fix string parsing of AoE Curse nodes

Nostrademous:
Adds support for parsing the following nodes on passive tree:

- Zone of Control
- Curse Area


LocalIdentity:
Fixed in https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/577

PR #589, opened 2025-01-24
Grant Skills from node / item now can add support gem

justjuangui:
Grant skill from nodes:
image

Grant skill from item:
image

Gems:
image

This PR also introduce the concept of "Set" with the following rule:
- if Set1 and Set2 then WeaponSet1 = true
- If not Set1 and Set2 then WeaponSet2 = true

Now to precalculate by skill we are taking into account if the gem is from item and if slot have swap to change the option

Triggered Skills isnt yet implemented in POB2

### DIFF
--- a/src/Classes/SkillsTab.lua
+++ b/src/Classes/SkillsTab.lua
@@ -169,11 +169,19 @@ local SkillsTabClass = newClass("SkillsTab", "UndoHandler", "ControlHost", "Cont
 		self.build.buildFlag = true
 	end)
 
+	self.controls.set1Enabled.enabled = function()
+		return self.displayGroup.source == nil
+	end
+
 	self.controls.set2Enabled = new("CheckBoxControl", { "LEFT", self.controls.set1Enabled, "RIGHT" }, { 50, 0, 20 }, "Set 2:", function(state)
 		self.displayGroup.set2 = state
 		self:AddUndoState()
 		self.build.buildFlag = true
 	end)
+
+	self.controls.set2Enabled.enabled = function()
+		return self.displayGroup.source == nil
+	end
 
 	self.controls.groupEnabled = new("CheckBoxControl", { "LEFT", self.controls.set2Enabled, "RIGHT" }, { 70, 0, 20 }, "Enabled:", function(state)
 		self.displayGroup.enabled = state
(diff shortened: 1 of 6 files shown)


src/Classes/SkillsTab.lua
```diff
@@ -208,41 +206,6 @@ local SkillsTabClass = newClass("SkillsTab", "UndoHandler", "ControlHost", "Cont
 	self.controls.groupCount.shown = function()
 		return self.displayGroup.source ~= nil
 	end
-	self.controls.sourceNote = new("LabelControl", { "TOPLEFT", self.controls.groupSlotLabel, "TOPLEFT" }, { 0, 30, 0, 16 })
```
Wires77, line 211: Something funny happens when I allocate Explosive Impact on the tree.  I'm assuming it's relating to removing this section.  Do we need to remove it?
image

src/Classes/SkillsTab.lua
```diff
@@ -1097,9 +1081,19 @@ function SkillsTabClass:SetDisplayGroup(socketGroup)
...
-		self.controls.groupSlot:SelByValue(socketGroup.slot, "slotName")
 		self.controls.groupEnabled.state = socketGroup.enabled
 		self.controls.includeInFullDPS.state = socketGroup.includeInFullDPS and socketGroup.enabled
+		if socketGroup.sourceItem then
+			local swap = not not socketGroup.slot:find("Swap")
+			socketGroup.set1 = not swap
+			socketGroup.set2 = swap
+		elseif socketGroup.sourceNode then
+			socketGroup.set1 = true
+			socketGroup.set2 = true
```
Wires77, line 1092: Can you not set a skill from a node to a weapon slot in-game?

src/Modules/CalcSetup.lua
```diff
@@ -175,7 +175,7 @@ function calcs.buildModListForNode(env, node, incSmallPassiveSkill)
 			t_insert(node.grantedSkills, {
 				skillId = skill.skillId,
 				level = skill.level,
-				noSupports = true,
+				noSupports = skill.noSupports ~= nil and skill.noSupports,
```
Wires77, line 178: Can't you just do `skill.noSupports`?  That's what is [in Item.lua](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/dev/src/Classes/Item.lua#L1751)

Wires77 (commented):
I'm still doing some testing, but I found a couple of oddities by just reading the code so far

src/Classes/SkillsTab.lua
justjuangui: Thx i didnt realice that kind of source, fixed in the last code

src/Classes/SkillsTab.lua
justjuangui: can you elaborate more please, i would like to understand your point

src/Modules/CalcSetup.lua
justjuangui: Done

Alexander3a:
just asking is there any updates on this?
i would love to finnaly have this feature 
(cant even get it to correctly work in my fork anymore)

majochem:
So I tested with this PR a little bit when working on attribute stuff and the Side Panel seems to correctly show the damage and gem attribute requirements of the selected skill based on the weapon set it is assigned to, but I noticed the following:

- All non-skill related values in the side panel like HP values, Attributes, or Attribute Requirements from Items/Tree are still based on the active set that is set on the items tab, even if Active skill cannot be used with that weapon set

- The effect of weapon-set specific passive is correctly taken into account for the DPS calculation, however the hover compare tooltip for passive nodes is again still based on the active item set regardless of skill weapon set assignment

PR #603, opened 2025-01-25
Add support for Inevitable Critical support gem

### LINKED ISSUE #298: Add support for Inevitable Critical Support
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### Check for support

Windows

- [Inevitable Critical Gem] In game when I use Falling Thunder and I haven't crit in 4 seconds it boosts my crit chance by 100%. Skill breakdown in-game shows 40.48%.
- [Longshot Gem] In game when I use Falling Thunder and my enemy is further than 7 meters it increases my attack damage by 35%

- [Inevitable Critical Gem] In PoB there doesn't to appear to be such a calculation, my projectile crit chance for FT says 31.68% even though I don't have "Have crit recently" checked in config. 
- [Longshot Gem] Even though I've set my config to 7 for "Melee distance to enemy" (I'm assuming meters is the unit) the damage doesn't change at all. (Also set projectile travel distance to 7 in case that's what it's looking at). 

1. Select Falling Thunder projectile as main skill.
2. Toggle "Have you Crit Recently"? on and off and see how Crit Chance doesn't change

1. Select Falling Thunder projectile as main skill.
2. Toggle "Melee distance to enemy"/"Projectile travel distance" on and off and see how skill attack damage doesn't change. 

### Character build code

```shell
https://maxroll.gg/poe2/pob/75qb0y0l
```

## Configuration
Image

## Crit Calc

Image

jjbi123:
Fixes #298 

- Inevitable Critical support gem had no support

### Before screenshot:
image

### After screenshot:
image

image


Blitz54:
Saw your comment on discord about capping it to 100. I think I got it figured out decently. 
`	{ var = "SecondsSinceInevitableCrit", type = "count", label = "# of seconds since Inevitable crit:", ifSkill = "Inevitable Critical", apply = function(val, modList, enemyModList)
		if val >= 4 then
			modList:NewMod("Multiplier:SecondsSinceInevitableCrit", "BASE", 4, "Config")
		else
			modList:NewMod("Multiplier:SecondsSinceInevitableCrit", "BASE", val, "Config")
		end
	end },`

I think this is a good way to go about it. We simply check if the value in the box is greater or equal to 4, and if so, we replace the value with 4. So you could have 45 in the box and it will send 4, which is the 100% increase.

src/Data/Skills/sup_int.lua
```diff
@@ -2060,6 +2060,14 @@ skills["SupportInevitableCriticalsPlayer"] = {
 			label = "Inevitable Critical",
 			incrementalEffectiveness = 0.092720001935959,
 			statDescriptionScope = "gem_stat_descriptions",
+			statMap = {
+				["support_inevitable_criticals_critical_strike_chance_+%_per_second"] = {
+					mod("CritChance", "INC", nil, 0, 0, { type = "Multiplier", var = "SecondsSinceInevitableCrit", limitVar = "InevitableCritCap" }),
```
PJacek, line 2065:
```suggestion
					mod("CritChance", "INC", nil, 0, 0, { type = "Multiplier", var = "SecondsSinceInevitableCrit", limitVar = "InevitableCritCap", limitTotal = true }),
```

src/Export/Skills/sup_int.txt
```diff
@@ -543,6 +543,14 @@ statMap = {
 #skill SupportInevitableCriticalsPlayer
 #startSets
 #set SupportInevitableCriticalsPlayer
+statMap = {
+	["support_inevitable_criticals_critical_strike_chance_+%_per_second"] = {
+		mod("CritChance", "INC", nil, 0, 0, { type = "Multiplier", var = "SecondsSinceInevitableCrit", limitVar = "InevitableCritCap" }),
```
PJacek, line 548:
```suggestion
		mod("CritChance", "INC", nil, 0, 0, { type = "Multiplier", var = "SecondsSinceInevitableCrit", limitVar = "InevitableCritCap", limitTotal = true }),
```

PJacek (changes requested):
You should use `limitTotal`. It makes the limit apply to the total value after multiplication.
I'm not convicted by this implementation. Do we really expect users to figure out if they crit 2 or 3 seconds ago?
In my opinion a checkbox would be a better solution. At least until someone good enough at maths figures out how to calculate the effective impact.

jjbi123:
> You should use `limitTotal`. It makes the limit apply to the total value after multiplication. I'm not convicted by this implementation. Do we really expect users to figure out if they crit 2 or 3 seconds ago? In my opinion a checkbox would be a better solution. At least until someone good enough at maths figures out how to calculate the effective impact.

thanks! limitTotal makes sense. Regarding the checkbox vs number I can only speak for myself as a monk but I would prefer the ability to see just how much my crit chance is affected by the second. For example, I have it on Falling Thunder, and when I throw down a fully "charged" attack at 4 seconds I know it'll always crit, but I also get power charges back from that hit, so I could immediately Falling Thunder again, and would be nice to know what my crit chance is, say 1-2 seconds after my last hit, or if I should just wait a bit to get it fully charged again. Idk, given that PoB is all about customization and nitty-gritty stuff this seems to be in line with the general philosophy to err on the side of giving users more control than abstracting functionality. Thoughts?

PR #637, opened 2025-01-27
Added Times Stunned Recently and edited Warcries used Recently

### LINKED ISSUE #573: Warcries used Recently and Stunned Recently value should default to 1 when flag set
### Check for duplicates

Windows

Was looking into "Briny Carapace" which increases stun threshold based on the # of times you were stunned recently, and while I did get it working, I think it could be better. I followed the same structure of "# of Warcries used recently" and my issue is that setting the checkbox for "used a warcry recently" only sets the flag, but I think it should also set the # used to 1 as default, as it would be guaranteed you have used at least 1. This was more noticeable when I added "# stunned recently" because of the two small passive nodes before Briny Carapace, because they would lower your threshold by 40%, but the 60% increase wouldn't activate unless you manually entered a value into the box.

Warcry does the same, it sets the flag, but there is no DPS increase when using "Deafening Cries" unless you manually enter a value.

Both have the issue of unchecking the flag box but leaving a value in the text box.

The "# of warcries used recently" and "# of times stunned recently" (when  implemented) should default to 1 if the checkbox is ticked. I think it should also prevent the user from entering a value less than 1 if the checkbox is enabled. This would be much more user friendly for people new to PoB.

That, or remove the checkbox and have just the value box for setting the flag, but I think that is also unintuitive for the average user.

Currently you can just set the value box to whatever number you need.

### Additional context

Image

Blitz54:
Fixes #573

Number of times Stunned Recently implemented. Briny Carapace uses this, and Quick Recovery probably does? I assume it stacks, I haven't touched that anyway yet. Also changed warcries used recently to clean up config.

EDIT: ooooooh cant use the # sign at the start of a sentence, thats why my text was all wonky.

Short form

- Times Stunned Recently is hidden unless Stunned Recently is toggled on
- Same for # Warcries used Recently
- The default value in each number box is 1, it is guaranteed the player used at least 1 warcry if toggled
- This also fixes the issue of toggling warcries, but not having any damage increase from "Deafening Cries" unless you manually enter a value into the box.
- When unchecking "Stunned Recently", the # times box will go red, and the values inside the box are not calculated as we check for the status of the toggle box.
- Same for # of Warcries, before if you left a number in the box but turned off the toggle, you would still get the dps.

Please check it over, I think you will like

-
-
-

### After screenshot:
image
image

### DIFF
--- a/src/Data/ModCache.lua
+++ b/src/Data/ModCache.lua
@@ -1,4 +1,7 @@
-local c=...c["+0.15% to Thorns Critical Hit Chance"]={{[1]={flags=0,keywordFlags=0,name="CritChance",type="BASE",value=0.15}}," Thorns  "}
+local c=...c["+0% to Cold Resistance"]={{[1]={flags=0,keywordFlags=0,name="ColdResist",type="BASE",value=0}},nil}
+c["+0% to Fire Resistance"]={{[1]={flags=0,keywordFlags=0,name="FireResist",type="BASE",value=0}},nil}
+c["+0% to Lightning Resistance"]={{[1]={flags=0,keywordFlags=0,name="LightningResist",type="BASE",value=0}},nil}
+c["+0.15% to Thorns Critical Hit Chance"]={{[1]={flags=0,keywordFlags=0,name="CritChance",type="BASE",value=0.15}}," Thorns  "}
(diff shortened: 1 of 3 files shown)


Blitz54:
Also just checked, but should we add "Hit Recently" to Stunned recently? Keep both boxes of course, because there are times you are hit but not stunned. But if we got stunned, we were clearly hit.

LocalIdentity (approved):

PR #677, opened 2025-01-30
Added support for Crown of thorns 5 life on using skill

Blitz54:
Fixes # .

Unique helm Crown of Thorns has the mod "lose 5 life when you use a skill". I feel the best way to represent that is just to add base 5 life cost to all skills, other than persistent skills.

-Confirmed in game that activating persistent skills does not cost life.

https://maxroll.gg/poe2/pob/185h30ft

### After screenshot:
image


LocalIdentity (changes requested):
Life lost is not the same as life cost so we can't show it in the same way. We need to add wider support for things life life sacrificed for list lost to be included

LocalIdentity:
Not required until we support all kinds of life lost or life sacrificed mods

PR #679, opened 2025-01-30
Added support for Minions resistances are equal to yours

Blitz54:
Fixes # .

Carrion Call unique shield has the mod "minions' resistances are equal to yours". This works, but looking for clarification on stuff.

Not totally sure on exactly how it functions. As you can see, the minion gets my 76% resistance, even though it says MAX 75. I can't get that to function in pob without just overriding the max res to 76. Understandable though.

Other thing I'm unsure of, is whether the minion gets player capped resist, or overcapped resist. I see nowhere on the minion page to see the minions overcapped res, so I'm not sure how we can tell.

Also, the code is pretty straight forward, but not very compact. If there is a better way to do this let me know!

-
-
-

https://maxroll.gg/poe2/pob/hj5lf0fj
### After screenshot:
image
image


Blitz54:
Removed the maximum resistance changes to minions. As it is, it takes players capped resistances, and caps it to the minions max resistance.

E.g. Player max fire resistance is 76. Minions is 75, so minion stat shows 75% fire res with 1% overcap.

https://maxroll.gg/poe2/pob/bq82i01q

LocalIdentity (approved):

PR #680, opened 2025-01-31
Widened dropdown to properly display ascendancy class names

MrHB212:
Fixes # .

The dropdown box was too small to properly display some of the ascendancy class names.
-
-
-

### After screenshot:
after


Wires77 (approved):
This top bar will likely change since it already doesn't fit on narrow screens, but I'll merge this for now

PR #682, opened 2025-01-31
Flame Wall DPS was not showing

MrHB212:
Fixes # .

-
-
-

### Before screenshot:
before

### After screenshot:
after


LocalIdentity (changes requested):
The .lua files and their skill types do not get manually changed. They are pulled directly from the game fiels

Issue #190, opened 2025-01-17
Add support for Pathfinder Ascendancy


QuickStick123:
### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### What platform are you running Path of Building on?

Windows

### Is your feature request related to a problem?

![Image](https://github.com/user-attachments/assets/9422b3ae-8188-4bee-88e3-2c6cd0f5dfd8)

![Image](https://github.com/user-attachments/assets/2a6bf3a8-0ceb-4b95-9679-d1754303ea41)

### Describe the solution you'd like

There are only 2 notable stats that need implementing.
For the movement speed penalty see https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues/189.

Relentless Pursuit just seems to be your action speed cannot be reduced by debuffs and should be implemented as such, by differentiating the penalties. https://www.poe2wiki.net/wiki/Slow

Blitz54:
Both nodes have been implemented now.

PR #700, opened 2025-02-04
Add parsing for "Your speed is unaffected by Slows"

RealWhimsy:
Fixes part of #190

This adds parsing for "Your speed is unaffected by Slows" modifiers, such as on Pathfinder's "Relentless Pursuit" notable or the unique "Wanderlust" boots.

- Restarted PoB with rebuilt ModCache
- Lines in "Relentless Pursuit" and "Wanderlust" now parse correctly (blue)
- Added an Attack (Molten Blast) to my Character and noted default Attack Rate: **1.26**
- Attack Rate when "Are you chilled?" is active drops to **1.13**
- Allocating "Relentless Pursuit" bounces the attack rate back to **1.26**, ignoring the slow of the chill

https://pobb.in/uP9iiR1OtT2i

### Before screenshot:
image

### After screenshot:
image

### DIFF
--- a/src/Classes/ModStore.lua
+++ b/src/Classes/ModStore.lua
@@ -138,6 +138,24 @@ function ModStoreClass:Sum(modType, cfg, ...)
 	return self:SumInternal(self, modType, cfg, flags, keywordFlags, source, ...)
 end
 
+
+--- Returns the value of all positive modifiers to a mod added together, ignoring any negative modifiers.
+--- Works by creating a table using Tabulate and then filtering for positive values.
+---
+--- @param modType string # the mod type for which we want to create the table, e.g. "INC" or "MORE"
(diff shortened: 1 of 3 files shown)


RealWhimsy:
Initially I implemented this as "your action speed cannot be modified to below base value", which is not correct. This way a 30% speed increase and a 20% speed decrease would result in a net 10% speed increase.

However, since our speed is unaffected by slows, it should ignore the 20% speed decrease and result in an overall 30% speed increase. I added a second commit that correctly introduces this behavior.

Implementation was briefly talked about on Discord with @Wires77 and @PJacek 


**Example config:**
image


**With Relentless Pursuit:**
image


**Without Relentless Pursuit:**
image

LocalIdentity (approved):

PR #713, opened 2025-02-05
Update Exposure Values in ModParser

IllusionElements:
Updated the exposure values in the modparser to the poe2 value of 20%.
Also adjusted the ailment code to allow for the shock override to work, because for shock it kept taking the minimum shock value, so if you entered in a larger value for the shock it would just take the default 20% but if you went lower it would take the lower value.
- ran the tests and ensured they all passed
- made sure pob-poe2 still worked after the changes and rebuilt the cache files
- tested both the exposure and shock value in the app, with the changes to ensure they were accurate

### DIFF
--- a/src/Modules/CalcPerform.lua
+++ b/src/Modules/CalcPerform.lua
@@ -1,9 +1,8 @@
 -- Path of Building
-
+--
 -- Module: Calc Perform
 -- Manages the offence/defence calculations.
 --
-
 local calcs = ...
 
 local pairs = pairs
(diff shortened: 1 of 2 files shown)


src/Modules/CalcPerform.lua
```diff
@@ -1,8 +1,10 @@
 -- Path of Building
---
+--package.cpath = package.cpath .. ";c:/Users/Aaron/.vscode/extensions/tangzx.emmylua-0.9.8-win32-x64/debugger/emmy/windows/x86/?.dll"
```
Blitz54, line 2: this should be removed

IllusionElements: Didn't realize it was still in there, removed it thanks

LocalIdentity (approved):

PR #741, opened 2025-02-08
Added support for ignore Warcry Cooldown

MrHB212:
### Before/after screenshots
cropped


src/Modules/ModParser.lua
```diff
@@ -2820,6 +2820,8 @@ local specialModList = {
 	["you are unaffected by bleeding while leeching"] = { mod("SelfBleedEffect", "MORE", -100, { type = "Condition", var = "Leeching" }) },
 	-- Warrior - Titan
 	["(%d+)%% increased effect of small passive skills"] = function(num) return { mod("SmallPassiveSkillEffect", "INC", num) } end,
+	-- Warrior - Warbringer
+	["ignore warcry cooldowns"] = function() return {mod("CooldownRecovery", "BASE", -8, { type = "SkillType", skillType = SkillType.Warcry })} end,
```
Wires77, line 2824: This should use a `MAX` of 0 instead of `BASE` in case more warcries are added with different base cooldown values.

src/Modules/ModParser.lua
```diff
@@ -3761,6 +3763,10 @@ local specialModList = {
 		mod("EnergyShieldRegen", "BASE", 1, { type = "Condition", var = "LifeRegenBurstFull" }, { type = "PercentStat", stat = "Evasion", percent = percent }),
 		mod("EnergyShieldRegen", "BASE", 1 / interval * duration, { type = "Condition", var = "LifeRegenBurstAvg" }, { type = "PercentStat", stat = "Evasion", percent = percent }),
 	} end,
+	["recover (%d+)%% of life and mana when you use a warcry"] = function(num) return {
+		mod("LifeRecovery", "BASE", 1, { type = "PercentStat", stat = "Life", percent = num}, { type = "Condition", var = "UsedWarcryRecently"}),
+		mod("ManaRecovery", "BASE", 1, { type = "PercentStat", stat = "Mana", percent = num}, { type = "Condition", var = "UsedWarcryRecently"}),
+	}end,
```
Wires77, line 3769: The reason the other recovery is over time is because the mana loss is also over time.  Have you tested this in-game to see?  We don't support generic instant recovery to my knowledge.

Wires77 (changes requested):

src/Modules/ModParser.lua
MrHB212: I have no means to test it. I think i will just remove it then, i shouldnt have added something i am not sure of. I will keep that in mind for the future.

MrHB212:
@Wires77 `MAX` didnt seem to work so i used `OVERRIDE` instead

LocalIdentity (approved):

PR #742, opened 2025-02-08
Added support for infernal flame

MrHB212:
- mana is replaced with infernal flame
- everything regarding mana is being hidden
- added checkboxes to config tab for "full infernal flame" and "high infernal flame"

Currently i am trying to implement the self inflicted damage when reaching full buildup and the buff when reaching the high status. 

### Before/after screenshot:
costs_cropped
configtab


MrHB212:
Here is how it looks like with the active nodes

final_cropped

PR #731, opened 2025-02-07
Add support for Armour Break

majochem:
Fixes #570 .

### Description of the problem being solved:
- Correctly identify conditions that allow Armour Break to be applied. Now supports:
    - Splinter Support
    - Devastate Support
    - Armour Breaker
    - Armour Piercing Rounds
    - "Shattering Blow" Notable 
    - Warbringer's "Anvil's Weight" Notable

- Add support for mods "% increased damage against enemies with Fully Broken Armour"
- Add option for partially broken armour in Config
- Display Armour values in the breakdown for Effective DPS of Physical Damage (will probably want to improve this further, but seemed very sensible for now to see effects)
- _Adds an `OVERRIDE` check to the `enemyArmour` calculation in `CalcOffence`, which is what I used to apply the "ArmourFullyBroken" values by overriding `Armour` to 0._

### Steps taken to verify a working solution:
- Check affected mod names are parsed correctly and highlighted "blue"
- Check Fully and Partially broken armour configs interact correctly 
    - custom value disables "ArmourFullyBroken" and associated bonuses
    - deleting custom value or setting to 0 re-enables fully broken armour
    - effective DPS values are calculated correctly
- Config option is highlighted red/invalid if none of the pre-requisites are fulfilled

### Link to a build that showcases this PR:
[...]

PR #754, opened 2025-02-10
Added support for "increased damage against fully armor broken enemies"

MrHB212:
### Before/after screenshot:
image


majochem:
Unless I'm missing something, I already added this as part of #731 , no?

MrHB212:
@majochem I am sorry, i am the one who was missing something.

PR #756, opened 2025-02-10
Added support for break armour on hit

MrHB212:
Currently it seems to work but i have to do some more testing. 

### After screenshot:
image

### DIFF
--- a/src/Modules/CalcOffence.lua
+++ b/src/Modules/CalcOffence.lua
@@ -347,7 +347,7 @@ function calcs.offence(env, actor, activeSkill)
 
 	-- Calculate armour break
 	local base = skillModList:Sum("BASE", nil, "ArmorBreakPerHit")
-	local inc = skillModList:Sum("INC", nil, "ArmourBreakPerHit")
+	local inc = skillModList:Sum("INC", nil, "ArmorBreakPerHit")
 	output.ArmorBreakPerHit = base * (1+inc/100)
 
 	local function calcAreaOfEffect(skillModList, skillCfg, skillData, skillFlags, output, breakdown)
--- a/src/Modules/CalcOffence.lua
+++ b/src/Modules/CalcOffence.lua
@@ -346,9 +346,9 @@ function calcs.offence(env, actor, activeSkill)
 	end
 
 	-- Calculate armour break
-	local base = skillModList:Sum("BASE", nil, "ArmorBreakPerHit")
-	local inc = skillModList:Sum("INC", nil, "ArmorBreakPerHit")
-	output.ArmorBreakPerHit = base * (1+inc/100)
+	local base = skillModList:Sum("BASE", nil, "ArmourBreakPerHit")
+	local inc = skillModList:Sum("INC", nil, "ArmourBreakPerHit")
+	output.ArmourBreakPerHit = base * (1+inc/100)
 
 	local function calcAreaOfEffect(skillModList, skillCfg, skillData, skillFlags, output, breakdown)
 		local incArea, moreArea = calcLib.mods(skillModList, skillCfg, "AreaOfEffect", "AreaOfEffectPrimary")
--- a/src/Modules/CalcOffence.lua
+++ b/src/Modules/CalcOffence.lua
@@ -345,6 +345,11 @@ function calcs.offence(env, actor, activeSkill)
 		return
 	end
 
+	-- Calculate armour break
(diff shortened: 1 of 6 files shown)


src/Modules/CalcSections.lua
```diff
@@ -624,6 +624,7 @@ return {
 	{ label = "ES % per second", color = colorCodes.ES, haveOutput = "ESPercentPerSecondHasCost", { format = "{2:output:ESPercentPerSecondCost}", { breakdown = "ESPercentPerSecondCost" }, { modName = { "ESCost", "Cost", "ESCostNoMult" }, cfg = "skill" }, }, },
 	{ label = "Rage Cost", color = colorCodes.RAGE, haveOutput = "RageHasCost", { format = "{0:output:RageCost}", { breakdown = "RageCost" }, { modName = { "RageCost", "Cost", "RageNoMult" }, cfg = "skill" }, }, },
 	{ label = "Rage per second", color = colorCodes.RAGE, haveOutput = "RagePerSecondHasCost", { format = "{2:output:RagePerSecondCost}", { breakdown = "RagePerSecondCost" }, { modName = { "RageCost", "Cost", "RageNoMult" }, cfg = "skill" }, }, },
+	{ label = "Armourbreak per hit", haveOutput = "ArmourBreakPerHit", { format = "{0:output:ArmourBreakPerHit}", { modName = "ArmourBreakPerHit", modType = "BASE"} }, },
```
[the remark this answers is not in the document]
MrHB212: You mean the displayed text ?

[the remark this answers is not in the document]
Nightblade: Yeah "Armourbreak per hit", although now I'm thinking maybe it should be:  "Armour Break per hit" since it appears like that in-game?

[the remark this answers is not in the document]
MrHB212: The displayed text is too long, i will go with "Armour Break / hit"
image
image

MrHB212:
### Summary
- added section in calc tab to show armour break per hit
- added support for `Break X% increased Armour`
- added support for `Break X% increased Armour on Enemies affected by Ailments`
- added support for `apply_X_armour_break_on_hit`

In the calc tab it looks like this:
image

As far as i can tell it works as it should.

src/Modules/ModParser.lua
```diff
@@ -3639,6 +3639,8 @@ local specialModList = {
 	["chaos damage with hits is lucky"] = { flag("ChaosLuckyHits") },
 	["elemental damage with hits is lucky while you are shocked"] = { flag("ElementalLuckHits", { type = "Condition", var = "Shocked" }) },
 	["break (%d+)%% of armour on heavy stunning an enemy"] = { flag("Condition:CanArmourBreak", { type = "GlobalEffect", effectType = "Buff", effectName = "ArmourBreak" }) },
+	["break (%d+)%% increased armour"] = function(num) return { mod("ArmourBreakPerHit", "INC", num)} end,
+	["break (%d+)%% increased armour on enemies affected by ailments"] = function(num) return { mod("ArmourBreakPerHit", "INC", num, { type = "Condition", var = "AppliedAilmentsRecently"})} end,
```
majochem, line 3643: I don't think this should be "AppliedAilmentsRecently" ?
The wording is "enemies affected by ailments", whether or not it was "applied recently" is not the deciding factor. It is whether there is any ailment on the enemy right now.

It should be possible to do this without a seperate config option as there are already existing conditions for ailments.

src/Modules/CalcOffence.lua
```diff
@@ -345,6 +345,11 @@ function calcs.offence(env, actor, activeSkill)
 		return
 	end
 
+	-- Calculate armour break
+	local base = skillModList:Sum("BASE", nil, "ArmourBreakPerHit")
+	local inc = skillModList:Sum("INC", nil, "ArmourBreakPerHit")
```
majochem, line 350: I think you can use `calcLib.val()` for this to do it all in one step? Either way, this current iteration would not take "MORE" multipliers into account

src/Modules/ModParser.lua
MrHB212: Youre right. Looking back i also wonder why i did it like that. I just changed it.

src/Modules/CalcOffence.lua
MrHB212: I was unaware of the function. Thanks.

LocalIdentity (approved):

PR #775, opened 2025-02-12
Add support for Candlemaker "chance to ignite is doubled"

Blitz54:
Fixes # .

Candlemaker unique gloves have "chance to ignite is doubled". Everything else that I see that is "doubled" uses a 100% MORE mod. So that's what I did here.

-

https://maxroll.gg/poe2/pob/sf3d800z
### After screenshot:
image


LocalIdentity (approved):

majochem:
Not important for now, but this mod, as well as all the other "doubled" modifiers might need to work via either multipliers with a limit and/or some kind of flags to prevent them applying multiple times. The specific "double" or "triple" wording usually implies that they are not multiplicative with other sources and only the highest applies. I don't think there's currently more than one source of "doubled" for all the affected stats in PoE2, but if there was, they'd multiply in PoB if you had more than one. (can easily test this by entering `Ignite Chance is doubled` in the config more than once.

Just thought I'd mention it, in case we see more "doubled" or even "tripled" mods in the future

PR #618, opened 2025-01-26
New culling strike logic, add rarity dropdown

pauloday:
Fixes #93 

### Description of the problem being solved:
Currently culling strike is still calculated as 10% for everything. Also culling strike threshold increases don't do anything. This PR fixes both issues, and adds a monster rarity dropdown in the config tab with the various default cull thresholds implement as well.

### Steps taken to verify a working solution:
- Ensure "Culling Strike" row in "Other Effects" table on calcs is 6.25%
- Unallocate "Cull the Hordes" on the passive tree
- Verify "Culling Strike" is now 5% in calcs tab
- Switch "Is the enemy a Boss?" to "No" in config tab
- Ensure monster rarity dropdown appears
- Select each option in the dropdown and check "Culling Strike" with and without "Cull the Hordes"
    - Normal should be 30%/37.5%
    - Magic should be 20%/25%
    - Rare should be 10%/12.5%
    - Unique should be 5%/6.25%

### Link to a build that showcases this PR:

```
[...]

PR #777, opened 2025-02-13
Add support for Attrition

Blitz54:

1% more damage per 2 seconds the enemy is in your presence properly works. And it limited to the gems maximum damage.

Thanks to @Paliak, thresholdVar works properly for the culling strike mod. Culling strike is enabled once the enemy is in your presence for the required time.

Culling Strike code will need to be changed slightly if #618 gets added, quick fix at that point.

### After screenshot:
image
image
image

### DIFF
--- a/src/Classes/ModStore.lua
+++ b/src/Classes/ModStore.lua
@@ -363,6 +363,14 @@ function ModStoreClass:EvalMod(mod, cfg)
 			end
 		elseif tag.type == "MultiplierThreshold" then
 			local target = self
+			local thresholdTarget = self
+			if tag.thresholdActor then
+			    if self.actor[tag.thresholdActor] then
+			        thresholdTarget = self.actor[tag.thresholdActor].modDB
+			    else
+			        return
+			    end
+			end
 			if tag.actor then
 				if self.actor[tag.actor] then
 					target = self.actor[tag.actor].modDB
@@ -378,7 +386,7 @@ function ModStoreClass:EvalMod(mod, cfg)
 			else
 				mult = target:GetMultiplier(tag.var, cfg)
 			end
-			local threshold = tag.threshold or target:GetMultiplier(tag.thresholdVar, cfg)
+			local threshold = tag.threshold or thresholdTarget:GetMultiplier(tag.thresholdVar, cfg)
 			if (tag.upper and mult > threshold) or (not tag.upper and mult < threshold) then
 				return
 			end
--- a/src/Classes/ModStore.lua
+++ b/src/Classes/ModStore.lua
@@ -365,11 +365,11 @@ function ModStoreClass:EvalMod(mod, cfg)
 			local target = self
 			local thresholdTarget = self
 			if tag.thresholdActor then
-			    if self.actor[tag.thresholdActor] then
-			        thresholdTarget = self.actor[tag.thresholdActor].modDB
-			    else
-			        return
-			    end
+				if self.actor[tag.thresholdActor] then
+			    	thresholdTarget = self.actor[tag.thresholdActor].modDB
+				else
+					return
+				end
 			end
 			if tag.actor then
 				if self.actor[tag.actor] then
--- a/src/Classes/ModStore.lua
+++ b/src/Classes/ModStore.lua
@@ -366,7 +366,7 @@ function ModStoreClass:EvalMod(mod, cfg)
 			local thresholdTarget = self
 			if tag.thresholdActor then
 				if self.actor[tag.thresholdActor] then
-			    	thresholdTarget = self.actor[tag.thresholdActor].modDB
+					thresholdTarget = self.actor[tag.thresholdActor].modDB
 				else
 					return
 				end
--- a/src/Data/Skills/act_str.lua
+++ b/src/Data/Skills/act_str.lua
@@ -762,6 +762,19 @@ skills["AttritionPlayer"] = {
 			label = "Attrition",
 			incrementalEffectiveness = 0.054999999701977,
 			statDescriptionScope = "attrition",
+			statMap = {
+				["skill_attrition_presence_max_seconds"] = {
+					mod("Multiplier:AttritionMaxDamage", "BASE", nil, 0, 0, { type = "GlobalEffect", effectType = "Buff"}),
+				},
+				["skill_attrition_culling_strike_at_x_or_more_stacks"] = {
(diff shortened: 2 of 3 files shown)


Blitz54:
Still can't get thresholdVar to work. Sigil of Power works properly when copying the PoB1 code. I think the issue might be because the thresholdVar for Attrition is a variable stat which changes with the skill gem level, where as the sigil of power max stages is a constantStat. I've tried everything else I can think of, not sure where to go from here.

Blitz54:
> I'll take a stab at the debugging later if you don't get around to it.
Much appreciated if you could. I took a look with the debugger, but it's a little beyond my skills at the moment. I'm really not sure why it won't work for me, tried everything I could think of and it just doesn't seem to operate as I expect, maybe I'm doing something horribly wrong but I really didn't think so.

I'll take a look at the git rebase you mentioned, haven't used git command lines much but I'll give it a shot.

Paliak:
Main issue you're running into here is that `MultiplierThreshold` uses the `actor` tag value for both the threshold value and the mult value. In this case the mult value is on the enemy actor (`EnemyPresenceSeconds`) and the threshold value is on player (`AttritionCullSeconds`). Because `AttritionCullSeconds` does not exist in the enemy mod db you always get 0.

Adding the ability to select the threshold target actor should fix the issue:

```diff
diff --git a/src/Classes/ModStore.lua b/src/Classes/ModStore.lua
index 7347ba64..e1fd8260 100644
--- a/src/Classes/ModStore.lua
+++ b/src/Classes/ModStore.lua
@@ -378,7 +378,15 @@ function ModStoreClass:EvalMod(mod, cfg)
            else
                mult = target:GetMultiplier(tag.var, cfg)
            end
-           local threshold = tag.threshold or target:GetMultiplier(tag.thresholdVar, cfg)
+           local thresholdTarget = target
+           if tag.thresholdActor then
+               if self.actor[tag.thresholdActor] then
+                   thresholdTarget = self.actor[tag.thresholdActor].modDB
+               else
+                   return
+               end
+           end
+           local threshold = tag.threshold or thresholdTarget:GetMultiplier(tag.thresholdVar, cfg)
            if (tag.upper and mult > threshold) or (not tag.upper and mult < threshold) then
                return
            end
diff --git a/src/Data/Skills/act_str.lua b/src/Data/Skills/act_str.lua
index 1745ae0e..4890e063 100644
--- a/src/Data/Skills/act_str.lua
+++ b/src/Data/Skills/act_str.lua
@@ -772,7 +772,7 @@ skills["AttritionPlayer"] = {
                },
                ["skill_attrition_hit_damage_+%_final_vs_rare_or_unique_enemy_per_second_ever_in_presence_up_to_max"] = {
                    {mod("Damage", "MORE", nil, 0, KeywordFlag.Hit, { type = "GlobalEffect", effectType = "Buff"}, { type = "Multiplier", var = "EnemyPresenceSeconds", actor = "enemy", limitVar = "AttritionMaxDamage", div = 2, limitTotal = true }, { type = "ActorCondition", actor = "enemy", var = "RareOrUnique" })},
-                   {mod("CullPercent", "MAX", nil, 0, 0, { type = "GlobalEffect", effectType = "Buff"}, { type = "MultiplierThreshold", var = "EnemyPresenceSeconds", actor = "enemy", thresholdVar = "AttritionCullSeconds"}, { type = "ActorCondition", actor = "enemy", var = "RareOrUnique" }),
+                   {mod("CullPercent", "MAX", nil, 0, 0, { type = "GlobalEffect", effectType = "Buff"}, { type = "MultiplierThreshold", var = "EnemyPresenceSeconds", actor = "enemy", thresholdActor = "player", thresholdVar = "AttritionCullSeconds"}, { type = "ActorCondition", actor = "enemy", var = "RareOrUnique" }),
                    value = 10,}
                    },
            },
```

To cleanup the git history all you need to do is to run `git rebase -i HEAD~10` (10 is the number of latest commits to show/work with) and then move the commits around (remember it goes top down, latest commits are at the bottom) and use the fixup commands to squash your misc commits into one, more coherent one (fixup squashes the commit with the commit above it).

Before messing with this though i'd recommend keeping a copy of the commits either here on github or somewhere else locally like on a different branch as it's fairly easy to remove commits entirely with git rebase.

Blitz54:
Thanks for looking into this! I assumed thresholdVar would operate the same as LimitVar in the damage mod directly above, as it seems to use the player actor regardless of setting `actor = "enemy"`

Managed to get the commits merged to the first initial one, it was a struggle. In the end I needed to force push, now I kinda understand... kinda. Made sure I backed it up before doing anything as you said.

I'll check out your code and add it in soon. Much appreciated for the help.

Blitz54:
@Paliak I made a small change, put the code in the same order as the "Multiplier" code. And I also set `thresholdTarget = self` instead of `thresholdTarget = target` for consistency. LimitVar uses the player actor as default unless stated otherwise, so I figure it makes sense to have MultiplierThreshold follow the same logic.

src/Classes/ModStore.lua
```diff
@@ -363,6 +363,14 @@ function ModStoreClass:EvalMod(mod, cfg)
 			end
 		elseif tag.type == "MultiplierThreshold" then
 			local target = self
+			local thresholdTarget = self
+			if tag.thresholdActor then
+			    if self.actor[tag.thresholdActor] then
+			        thresholdTarget = self.actor[tag.thresholdActor].modDB
+			    else
+			        return
+			    end
```
Nightblade, line 372: Hi, I see there's some mixed indentation here.

Blitz54: Oops, my apologies, I didn't catch that. Good now

src/Classes/ModStore.lua
```diff
@@ -365,11 +365,11 @@ function ModStoreClass:EvalMod(mod, cfg)
 			local target = self
 			local thresholdTarget = self
 			if tag.thresholdActor then
-			    if self.actor[tag.thresholdActor] then
-			        thresholdTarget = self.actor[tag.thresholdActor].modDB
-			    else
-			        return
-			    end
+				if self.actor[tag.thresholdActor] then
+			    	thresholdTarget = self.actor[tag.thresholdActor].modDB
```
Nightblade, line 369: I think you missed this :)

Blitz54: Bah. Fixed too. :facepalm:

LocalIdentity (approved):

PR #826, opened 2025-02-27
Implement Rare Templates

Blitz54:
Rare Templates were disabled for pob2 because of all the changes to metadata and basetypes and whatnot. I used poe2db to get the new metadata, and followed the old pob1 code for the different basetypes, and creating subtypes for the 'setBestBase' function. So now body armour, helmets, gloves, and boots have templates for the best item of that subtype in game. Subtype being Armour, Armour/Evasion, Evasion, etc.

I added a small check to ItemDBControl.lua for the Two Handed Melee option when searching both uniques and rare templates for base type. The problem was Two Handed Melee would include ALL staves. Not just melee, so spell based staves like "Dusk Vigil, Ashen Staff" would be in the list of melee. This is solved with that small change.

I also got the 'Shrine Sceptre' to show it's variants, as Purity of Lightning, Ice, and Fire are all options on that base name. I mostly copied the code for Dual Resist rings in PoB1.

Flails, Axes, Spears, and Swords have rare templates but are disabled for now. They still show up in the Crafting dropdown in the top right, but swords were doing that from the start. Spears and Flails are weird, everything functions but they don't grab the `#BaseGroup One Handed Physical One Handed` like maces, axes, and swords do. Not sure why, if I add the stats manually to each template then they work. Something to look into later, fine for now.

Replaced the dropdown select with 2 buttons. This way you only need to click once to switch, not twice. Please take a look if I could do it prettier. It feels weird to me, but it works and I don't wanna break it.

Also couldn't get flails to "unpack" properly with AbstractFlail. Again, no idea why. This works for now, I'd just like to understand WHY it wouldn't work the original way.

Any suggestions on rare templates let me know, I tried to follow PoB1. Otherwise this is basically done.

EDIT: For shields, the rare template for Best Base Evasion is a buckler that isn't in the game yet. I could remove that template for now, as the only pure evasion shields are bucklers anyway.

### Before screenshot:
image

### After screenshot:
image
image
image
image


Blitz54:
Flails aren't working for me, will need to take a closer look later. But basically, for whatever reason the flail.txt won't find them using '#baseMatch BaseType Metadata/Items/Weapons/OneHandWeapons/Flails/AbstractFlail' but if I switch it to '#baseMatch Metadata/Items/Weapons/OneHandWeapons/Flails/FourFlail%d+' it will find the flails and update flail.lua. But then any time I hover over a flail, be it a template or over the crafting dropdown in the top right, it throws an error.
image
I'm guessing it's the 'BaseType' part? Everywhere I looked seemed to have flail included, so not sure. Like I said I'll take a closer look, but if someone knows why right off the bat that would be great.

Blitz54:
Apologies for commits, still learning how it all works, using ammend commit in VS Code didn't operate how I expected it to from what I read.

LocalIdentity (approved):

PR #867, opened 2025-04-03
Toggle Attribute Nodes with Right Click

Kylixen:
The goal here was to make it easy to toggle set attribute nodes between the attributes with just a mouse click. If an attribute node is unset, then it will use the last used attribute to set it. If it's already set, it will toggle between attributes. This keeps state with the spec.attributeIndex so it behaves the same as using a hotkey to set the attribute.

- Fix contributions notes saying to use the x86 to x64 for debugging in JetBrains products
- enable a toggle-like effect when right-clicking an attribute node
- update modal dialogue's message to described the new right-click behavior


LocalIdentity (approved):

LocalIdentity:
Nice first PR, thanks for cleaning up the docs too!

PR #872, opened 2025-04-04
Passive Tree pre-update for version 0_2

justjuangui:
This PR enabled version "0_2" for creating.

this PR is only for prepare for launch


justjuangui:
^0Error loading main script: HeadlessWrapper.lua:121: LoadModule() error loading 'TreeData/0_2/tree.lua': cannot open TreeData/0_2/tree.lua: No such file or directory

While the new Tree isnt launch we are going to have this error

Wires77:
Something is off with the placement of Warbringer, Amazon, Ritualist, and Tactitian nodes, but I'm going to merge this to get a release going soon and we can fix it once we find the cause.

PR #873, opened 2025-04-04
pinned enemies cant evade

lucs66:
Fixes # .

Pinned targets can no longer Evade.

- tested by running the client

N/A

### Before screenshot:
image

### After screenshot:
image
image


lucs66:
Did not add light stun from the patch note as not sure how to implement that. Here is the full patch note:
Pinned targets can no longer Evade. Pinning a target now also causes them to be Light Stunned

PR #875, opened 2025-04-04
pinned enemies have no evasion

lucs66:
Fixes # .

Pinned targets can no longer Evade.

- tested by running the client

N/A

### Before screenshot:
image

### After screenshot:
image
image

### DIFF
--- a/src/Modules/CalcOffence.lua
+++ b/src/Modules/CalcOffence.lua
@@ -2181,8 +2181,7 @@ function calcs.offence(env, actor, activeSkill)
 				}
 			end
 		end
-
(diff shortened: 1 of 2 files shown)


lucs66:
attempted config change with 2nd commit, broke and now hit chance does not show properly. first commit works, but not cleanest solution.

Wires77:
You need to remove the `Condition:` in front of `CannotEvade` for this to work.  Also that config box doesn't seem to show up for anything, like Pin Support, for example.

LocalIdentity (approved):
The config box needs to show up if you have any source of pin in the build. Until we've added support for all those mods, we'll just permanently show it

PR #932, opened 2025-04-08
Add support for "Penetrate" (Amazon Ascendancy Notable)

majochem:
- Add support for "Penetrate" (Amazon Ascendancy Notable)
    - "Attacks using your Weapons have Added Physical Damage equal to 25% of the Accuracy Rating on the Weapon"
- Add support for generic "Attacks using your Weapons" modifiers

_Some enabling changes that are identical for this PR and PR #931_
- Adds local `isMartialWeapon` function to `CalcOffence`
- Adds certain weapon base stats to `output` in `CalcOffence` (base damage and accuracy for now) for use with "PerStat" mods
- Modifies `ModDBClass:SumInternal` to allow matching of sources that contain multiple ":" like e.g. "Item:3:Tactician, Gold Amulet"

- Mod gets parsed
- Damage values are added correctly and also take mods from runes into account
- Global Accuracy mods are not counted
- Works with different weapon types (only martial weapons)
- Distinguishes between main hand and off-hand weapon
- Only works for attacks that use your weapons (e.g. no Herald of Blood Explosion)

### Limitations:
- Added damage only shows up in the MH/OH Hit Damage breakdown (without source info) and not in the overall "Added Min/Max" breakdown, but changing that is out of scope for this PR

https://maxroll.gg/poe2/pob/pvghz0is

### After screenshot:
image
image

### DIFF
--- a/src/Classes/ModDB.lua
+++ b/src/Classes/ModDB.lua
@@ -96,7 +96,7 @@ function ModDBClass:SumInternal(context, modType, cfg, flags, keywordFlags, sour
 		if modList then
 			for i = 1, #modList do
 				local mod = modList[i]
-				if mod.type == modType and band(flags, mod.flags) == mod.flags and MatchKeywordFlags(keywordFlags, mod.keywordFlags) and (not source or ( mod.source and mod.source:match("[^:]+") == source )) then
+				if mod.type == modType and band(flags, mod.flags) == mod.flags and MatchKeywordFlags(keywordFlags, mod.keywordFlags) and (not source or ( mod.source and (mod.source:match("[^:]+") == source or mod.source == source))) then
 					if mod[1] then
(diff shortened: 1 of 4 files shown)


LocalIdentity (approved):

Vansaska:
Are we sure this consistently works? I've tried activating the Penetrate node on a couple of different characters and the DPS calculation never changes.

majochem:
> Are we sure this consistently works? I've tried activating the Penetrate node on a couple of different characters and the DPS calculation never changes.

Do you have a concrete example of a build where it doesn't work? (Might be better to open an issue for that, actually but I could still take a look)

Edit: @Vansaska I did some brief basic tests and it worked in those scenarios (local accuracy, accuracy runes, 1h, 2h). Happy to look into it if you provide a build code though

PR #955, opened 2025-04-09
Add support for Gemling Legionnaire's Integrated Efficiency node

LocalIdentity:
Hovering over support gems in the skillts tab does not show the DPS, but once socketed, the damage is reflected in the build


DavidBruun:
Hello. It seems like you've made this ascendancy global, so every support skill contributes to full dps. This is not correct, only the support skills attached to the actual skill gem, contributes to dps.
Right now picking this ascendancy in POB more than doubles my damage.

AkiTsugi:
Correct, what he said, atm, every single support gem in the character is giving stats. I wish it worked that way but it's only for the slotted skill rather than global.

Issue #992, opened 2025-04-13
PoB2 displays the wrong runes when pasting an item with multiple different socketed runes


Char7otte:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### What platform are you running Path of Building on?

Windows

### How is Path of Building expected to behave?

When pasting an item with multiple different runes socketed into the 'Items' tab, the dropdowns for socketing runes should correctly display the runes socketed.

### How does Path of Building behave?

Instead, there is a visual bug where the dropdowns select the wrong rune. This does not affect the actual stats of the item, the item preview and stat comparison are still correct. 

## Steps to Reproduce
1. Copy an item with two different socketed runes in POE 2.
2. Paste it into PoB2's **Item Tab**.
3. Check the socket dropdown for incorrect modifiers.

---

## Items Tested

| Item Type       | Bug Description                                  | Actual Runes Socketed          |
| ----------- |------------------------------------------------|-------------------------|
| Body Armour  | Shows `+12 Lightning Res` and `+10 Lightning Res`          | Storm Rune + Iron Rune  |
| Crossbow      | Shows `+110 Accuracy Rating` and `+80 Accuracy Rating`       |  Lesser Vision Rune + Vision Rune          |
| Crossbow      | Shows `+80 Accuracy Rating` and `+50 Accuracy Rating`      | Lesser Vision Rune + Lesser Resolve Rune |

---

### Bug Examples

<details>
<summary>⚠️ Body Armor with Storm Rune & Iron Rune socketed</summary>

```
Item Class: Body Armours
Rarity: Rare
Honour Shroud
Hawker's Jacket
--------
Quality: +20% (augmented)
Evasion Rating: 809 (augmented)
Energy Shield: 310 (augmented)
--------
Requires: Level 63, 70 Dex, 70 Int
--------
Sockets: S S 
--------
Item Level: 79
--------
+12% to Lightning Resistance (rune)
20% increased Armour, Evasion and Energy Shield (rune)
--------
+159 to Evasion Rating
+61 to maximum Energy Shield
[...]

PR #1010, opened 2025-04-15
Add proper support for pasting items with lesser and greater runes

QuickStick123:
Fixes https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues/992

As there are now multiple runes of different types it becomes harder to guess the runes used on the item.

This PR aims to solve this issue by generating a minimal combination of runes that meets the stats present on the item. This means it will always generate a set of runes that add to the total if it can, otherwise it won't. This does however mean some runes combinations won't be accurately reflexed as they are ingame e.g. 
image
Will get imported as 
image
But I think this is a minor issue as it should be right in most cases of people using the runes. But will sometimes get it wrong mostly with runes values that linear combinations of others e.g. if some uses 2 lesser +20 life runes it will always be importer as a single greater +40 life rune.

This approach uses a greedy base, and then performs search on top of it this should avoid efficiencies potentially lost due to inefficiencies, avoid searching duplicate combinations and is constrained as better searches are found. This was made using ChatGPT as a base them modifying it to suit the runes as well as a few optimisations and changes.

### Before screenshot:
433126833-1f6c73dd-ef82-4b28-b71f-27eb61a03130
433126839-70f077a5-680c-4dcc-ae31-337ae4134302
433126842-1b96938a-9221-4cf0-97ec-033d6c48d0fb

### After screenshot:
image
image
image

### DIFF
--- a/src/Classes/Item.lua
+++ b/src/Classes/Item.lua
@@ -853,13 +853,17 @@ function ItemClass:ParseRaw(raw, rarity, highQuality)
 								local function checkUnique(result)
 									-- Generate a unique key from the result table this prevents duplicates combinations being searched
 									local key = ""
-									for value, count in ipairs(result) do
+									for value, count in pairs(result) do
 										if count > 0 then
-											key = key .. value .. "x" .. count
+											key = key .. value .. "x" .. count .. " "
 										end
 									end
-									if visited[key] then return end
-									visited[key] = true
+									if visited[key] then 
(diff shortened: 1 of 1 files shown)


Wires77:
I don't have runes to test with in the latest patch, does this work properly when importing via the API still?

LocalIdentity (approved):

PR #1100, opened 2025-05-04
Add support for Forking Critical Strikes

### LINKED ISSUE #1081: Add Tangletongue crit “forking” to damage calculations
### Check for duplicates

Windows

Tangletongue doesn’t follow the standard crit formula. Given the immense popularity of this weapon, please consider trying to add support for it’s “forking” calculations within POB2.

I suspect the Tangletongue “forking” behaviour will change optimisation in its builds pretty significantly - crit subs are probably much more valuable than POB currently suggests (CR moreso than CD).

Thanks devs for your hard work.

Enable Tangletongue “forking” in damage calculations.

_No response_

### Additional context

_No response_

Antaresque:
Fixes #1081 .

As Tangletongue is one of most used uniques in current patch, it should be added as its addition would be very useful to most of players using PoB.
Forking crits, according to in-game descriptions and players' observation:
- rolls Critical Strike twice (similar to Lucky, but both works alongside each other),
- if only one of them rolls, then its normal crit,
- if both roll, then critical bonus is doubled.

Implementation adds flag CritFork from modifier, which have to support: 

#### Double roll on Critical Chance 
(same as already implemented Lucky critical, but after Lucky happens)
50% base chance -> 75% odds of hitting single critical strike with Forking
50% base chance -> 75% of hitting with Lucky -> 93,75% of hitting with Fork/Lucky
Lucky should be first as each Fork has separate Lucky roll if both modifiers are together.

#### Multiplying Crit Multplier on forking Crits 
if both rolls succeed, then multiplier is doubled:
e.g. 50% odds for fork -> half of crits are 300 multi, other half is 200 multi (assuming default multiplier)

Implementation calculates it using average Crit Multiplier (for example above, it would be 250):
(critchance before fork) * (critchance before fork) / (critchance before accuracy penalty) -> average MORE multiplier provided by forking, e.g. 50% base -> 25% for fork hit/25% more crit multiplier

Correction for accuracy must be made due to calculating only for critical hits (other dont have crit multiplier, so they're irrelevant for average), so avg. crit multplier stays same as intended, despite hit chance changing (and, due to it, effective crit chance).

#### Unresolved
- it may be useful to have displayed chance for double-criting in skill stats on left (as its in-game)
- unsure if critical multiplier should have "average" clause for clarity when fork flag is appearing
- implementation most likely could be cleaner, as currently there is both before-fork and before-accuracy critchance in output
- still need to test if everything works properly as I am not that familar with codebase at the moment

https://maxroll.gg/poe2/pob/4f59g0on

### After screenshot:
image
image


LocalIdentity (approved):
I made quite a few changes to the PR to fix some issues surrounding hit chance. I also added a breakdown for the chance to get a forked crit

PR #1106, opened 2025-05-06
Add support for some "enemies in your presence" mods

Blitz54:
Since we now have the presence area from #1094, we can now add proper checks for EnemyInPresence.
Had to change the calculation between presence and enemyDistance as it was backwards and setting it to true if the enemy distance was further than presence radius.

### Issue
Evergrasping Ring doesn't seem to do anything for EHP. Putting a custom mod of "Enemies in your presence deal 20% more chaos damage" affects EHP as expected. So I think it's just a problem with "gain as extra" not working on the enemy for some reason.


majochem:
> Had to change the calculation between presence and enemyDistance as it was backwards and setting it to true if the enemy distance was further than presence radius.

Oops 😬 
Thanks you for fixing that!

Apart from that. I think it might be good to still have a config option á la "Enemy is always in your presence" to be able to test effects of distance without "interference" from presence related mods. Thoughts?

LocalIdentity (approved):

PR #1115, opened 2025-05-07
Fix crash when deleting newly created config set

### LINKED ISSUE #1077: Creating a new set and clicking Rename, Copy, Delete crashes the client
### Check version

### Check for duplicates

Windows

### Screenshots of crash/graphical issue, version and options menu

Image

This crash has multiple different crash reports. I have included one because I think it would be too much and they are mostly the same. Manage Trees in the Tree tab does not have this issue.

### What are your system specifications and configuration?

Windows 11

-Create new build
-Select configuration, skills, or items tab
-Select Manage(skill set, item set, config set)
-Select new
-Type in anything and click save
-Do not click anything else
-Click Delete, Copy, or Rename

### PoB for PoE2 build code

```shell

```

Paliak:
Fixes #1077

Creating a new config set would save the entire table into self.selValue which is then later used as an index. Since the configSets table is indexed by ids instead it returns null and causes a crash.

This pr simply makes the save button function save the id instead of the full table. This also makes the newly created config set highlight nicely.

See reproduction steps in issue

### Before screenshot:

obraz

### After screenshot:

obraz

### DIFF
--- a/src/Classes/ConfigSetListControl.lua
+++ b/src/Classes/ConfigSetListControl.lua
@@ -52,7 +52,7 @@ function ConfigSetListClass:RenameSet(configSet, addOnName)
 		if addOnName then
 			t_insert(self.list, configSet.id)
 			self.selIndex = #self.list
-			self.selValue = configSet
+			self.selValue = configSet.id
 		end
 		self.configTab:AddUndoState()
 		self.configTab.build:SyncLoadouts()


LocalIdentity (approved):

Phobic-HR:
Crash is still reproducible in two other locations. Manage Skill Sets and Manage Items Sets.
Line 65 in SkillSetListControl and Line 56 in ItemSetListControl.

PR #1190, opened 2025-08-26
Fix vertical spacing and tooltip header size

Blitz54:
Removed a 2 pixel offset for tooltips with pretty UI on. Originally it matched the total height of the OG tooltip, but it created uneven black space. Removing it evened out the black space, and shrinks total tooltip size a little.

Also stretched some headers to better fit letters like q and p that go below the base font line.

And shrunk the separator below corrupted as it was slightly bigger than the rest.


Nightblade (approved):
LGTM

LocalIdentity (approved):

PR #1230, opened 2025-08-29
Update Witch Hunter nodes to 0.3

ButterRangers:
Fixes # .

The 0.3 patch updated two Witch Hunter ascendancy nodes:
Weapon Master was updated to now grant 100 weapon swap skill points rather than 20.
It's prerequisite node was updated to grant 6% increased cooldown recovery rate rather than an addition 4 weapon swap nodes.

This PR updates these values.
- Updated any mention of Weapon Master and its prerequisite node to feature the new stats.
- See screenshots.

I'm not sure what kind of build is meant. CI build?
### Before screenshot:
image

### After screenshot:
image


LocalIdentity:
The 0.3 tree will be exported directly from the game files when we have access to the data in 5 hours.

PR #1233, opened 2025-08-29
Fix: Prevent weapon set A node allocation in weapon set B path

TheDechev:
# Fix weapon set path crossing allocation

## Summary

Fixes a bug where users could allocate nodes across different weapon sets through path crossing. Users could allocate a node for weapon set 2 even if the only path went through nodes allocated to weapon set 1.

### Weapon Set Path Validation
• Added `GetNodeBlockingReason()` with detailed validation logic for different blocking scenarios
• Unified allocation/deallocation blocking in `shouldBlockNodeAllocation()` and `shouldBlockNodeDeallocation()` functions  
• Connection-based validation checks entire path connectivity including linked nodes

### Enhanced User Warnings
• Specific tooltip messages for different blocking reasons (global nodes, path crossing, path through globals)
• Integrated with existing global node warning system

### Optimized Validation Logic
• Efficient set-based path checking eliminates nested loops
• Unified direct link and path validation handling
• Directional weapon set compatibility (weapon sets can connect to main tree, not vice versa)

### Images
image
image

## Validations
• Tested weapon set switching and cross-weapon-set allocation attempts
• Verified far node allocation works correctly within same weapon set
• Confirmed blocking works for paths through global nodes while in weapon set mode
• Ensured tooltip warnings appear correctly with specific guidance

### DIFF
--- a/src/Classes/PassiveTreeView.lua
+++ b/src/Classes/PassiveTreeView.lua
@@ -168,11 +168,11 @@ function PassiveTreeViewClass:Draw(build, viewPort, inputEvents)
 	local offsetY = self.zoomY + viewPort.y + viewPort.height/2
 	local function treeToScreen(x, y)
 		return x * scale + offsetX,
-		       y * scale + offsetY
+			   y * scale + offsetY
 	end
 	local function screenToTree(x, y)
 		return (x - offsetX) / scale,
-		       (y - offsetY) / scale
+			   (y - offsetY) / scale
 	end
 
 	if IsKeyDown("SHIFT") then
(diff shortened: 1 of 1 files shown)


src/Classes/PassiveTreeView.lua
```diff
@@ -271,42 +271,42 @@ function PassiveTreeViewClass:Draw(build, viewPort, inputEvents)
...
-		local shouldBlock = weaponSetMode or connectedToWeaponSetNodes
+		local isGlobalNode = node.type == "Keystone" or node.type == "Socket" or node.containJewelSocket
 
-		return shouldBlock
+		if isGlobalNode then
+			local weaponSetMode = spec.allocMode > 0
+			local connectedToWeaponSetNodes = self:IsConnectedToWeaponSetNodes(node)
+			return weaponSetMode or connectedToWeaponSetNodes
+		else
+			return not self:PathCrossesWeaponSets(node, spec)
```
Wires77, line 287: This function doesn't exist, so I get an error when pathing past a jewel socket:
image

PR #1235, opened 2025-08-29
Export base items

Wires77:


src/Data/Bases/sword.lua
```diff
@@ -15,7 +15,6 @@ itemBases["Golden Blade"] = {
 itemBases["Energy Blade One Handed"] = {
 	type = "One Handed Sword",
 	quality = 20,
-	hidden = true,
```
Wires77, line 18: This is actually instance one of two in this file, so the second one (with hidden = true) overrides this one

LocalIdentity (approved):

PR #1239, opened 2025-08-30
Add support for adding essence modifiers to items

Wires77:
Mods fix in this PR is similar to the one in #1242, with this one being more sustainable.  I can fix the merge conflict if that one is merged first.

The bulk of the essence code is restored from #6 and adapted to the new spec


Blitz54:
The ModCorrupted.lua is borked. Not sure why exactly, the weightKey and weightVal are missing. Maybe from a recent Spec change?

Wires77:
> The ModCorrupted.lua is borked. Not sure why exactly, the weightKey and weightVal are missing. Maybe from a recent Spec change? Tried looking last night but didn't solve yet.

Looks like they added the corrupted mods to GoldModPrices, but they don't have spawn weights, so our fallback logic doesn't get a chance to fix that.

PR #1242, opened 2025-08-31
Update current Uniques and add 0.3.0 Uniques

Blitz54:
Updated all the current uniques. Lots are just wording changes that affect current items.

There is probably a better way to fix the mod exporter. I did what worked for now so I could get the uniques done, but it's probably done poorly.


src/Export/Scripts/flavourText.lua
```diff
@@ -31,6 +31,20 @@ local forcedNameMap = {
 	["FourUniqueSceptre6a"]  = "Guiding Palm of the Heart",
 	["FourUniqueSceptre6b"]  = "Guiding Palm of the Eye",
 	["FourUniqueSceptre6c"]  = "Guiding Palm of the Mind",
+	["FourUniqueBootsStrDex4"]  = "Shankgonne", -- 0.3.0 Uniques are missing from Unique Stash Layout.
```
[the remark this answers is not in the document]
Blitz54: > Oh no...this might not be sustainable if that's the case

I don't think it's too much of an issue. A little annoying yes, but doesn't take that long to do.

I did have a python script to get the flavour text from trade, but it's easier to use PoE2DB as they check trade anyway.

src/Data/Uniques/shield.lua
```diff
@@ -73,23 +81,27 @@ Recover 4% of maximum Life when you Block
...
 Effigial Tower Shield
+Source: No longer obtainable
 Implicits: 1
 Grants Skill: Raise Shield
 10% reduced Movement Speed
 (100-150)% increased Armour
 +(100-150) to Stun Threshold
-Can Block damage from all Hits
-Cannot use Shield Skills
+UniqueCanBlockAllDamage1
```
Wires77, line 104: Hmm, not sure what to do about mods like this that just get deleted

Blitz54: Oopsie, removed the wrong mod. Legacy versions still have the cannot use shield skills, they no longer have block from all hits. Will fix.

Issue #994, opened 2025-04-13
Cannot Authorise PoE with Wine (Linux)


michaldybczak:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### What platform are you running Path of Building on?

Linux - Wine

### What is the wording in-game?

Authorize process should work normally with Wine.

### What is the wording in Path of Building?

I tried to authorize app but when I click "Authorize" button on PoE site, it opens site:

http://localhost:49082/?code=1fb53e16c2a5bd552d327e80cb47b2240bb95230&state=88c2ec5e98f94320

Browser cannot open localhost:49082

I assume, that Wine somehow reroutes this and lands on localhost instead of the app, which breaks the process.
 

### PoB for PoE2 build code

```shell

```

michaldybczak:
It finally worked! I initiated it again and again from PoB2 and somehow it was successful. But on PoE site there was a message:

_We cannot verify that this request has come from Path of Building. You are responsible for any requests made using your account._ 

I thought that this won't work at all, but in general it worked but with some problems.

Wires77:
`localhost` is the expected URL here in order to let PoB know the authentication request succeeded.  There is a timeout of 30 seconds, so it's possible you ran into that the first time you tried.

That message on the Path of Exile website is also expected, and it's because there is no way to 100% guarantee PoB sent you to that site and not another application, so just make sure you initiated the flow from PoB if you see that unexpectedly.

zao:
There's two types of client for the OAuth API - "confidential client" and "public client".
The warning on the authorisation page always occur for public clients, which Path of Building is.

Most of the PoE-related websites and many apps you use around the game are confidential clients. They have a dedicated secure server that gets the tokens and performs actions on your behalf. Even if such apps seems to run locally on your computer, they have some sort of server backend.

Public clients is a newer kind of client that doesn't have this kind of server backend, they are solely a standalone program running on your computer. As there's no trusted server involved, all the authorisation server can work with is the registered application that the initiating program claims it is but there's no secure way to store a secret to prove this as everything runs on your computer.

PR #944, opened 2025-04-09
Fix "Could not generate search" error when using only full dps as a weight and not using full dps

Paliak:
Fixes issue mentioned on discord
Fixes #925
### Description of the problem being solved:

#8247 removed handling for this case. My intention was to make it so that when fulldps is the only weight and it is not set then it is effectively ignored but that caused this issue. This pr re-adds the if check that handles this situation as before.

### Steps taken to verify a working solution:
- Test by attempting to find best Weapon 1 in test build

### Link to a build that showcases this PR:
https://pobb.in/AGQ1ldvG_rUz

PR #1285, opened 2025-09-03
Fix OAuth2 redirect timeouts caused by closing the server too early

redgoldlace:
Fixes the issue mentioned in #994, where redirects don't behave properly.

I've explained it in greater detail in a comment in `LaunchServer.lua`, but this fixes a hard-to-pin-down timing issue that caused OAuth authorization to fail. More specifically, the server only waited for one connection before closing, so _anything_ elsewhere on the OS misbehaving (be it a loopback adaptor, a VPN/proxy application, or some other program) by opening a connection would cause the server to give up and close, _before_ the actual OAuth redirect reached it, which resulted in the OAuth redirect timing out, as described in #994. This prevented me from importing characters - since the OAuth redirect never completed, PoB never received the authorization code necessary to retrieve an access/refresh token later. I assume it's something many others have run into once or twice as well.

- Repeated testing in a development environment
- Static type-checking provided by EmmyLua

Environment-dependent, any build or account could cause this issue.

### Before screenshot:

image

### After screenshot:

image

### DIFF
--- a/src/LaunchServer.lua
+++ b/src/LaunchServer.lua
@@ -95,15 +95,15 @@ ConPrintf("Opening URL: %s", url)
 OpenURL(url)
 
 --- Handle an incoming socket connection, to complete an OAuth redirect.
---- @param client table The socket connection to handle, as returned by `server:accept()`.
---- @param attempt number The number of attempts made to handle an incoming connection. This is used for logging
+--- @param client table @The socket connection to handle, as returned by `server:accept()`.
+--- @param attempt number @The number of attempts made to handle an incoming connection. This is used for logging
 --- purposes, since spurious issues can be difficult to identify otherwise.
---- @return boolean shouldRetry Whether we should wait for another connection. If false, we've successfully responded to
---- a HTTP request. Note that, for the purposes of this function, we don't care whether authorization was *granted*,
+--- @return boolean shouldRetry @Whether we should wait for another connection. If false, we've successfully responded
+--- to a HTTP request. Note that, for the purposes of this function, we don't care whether authorization was *granted*,
 --- just that the process itself was completed and the user was redirected as intended.
---- @return string? code The OAuth authorization code. This is exchanged for an access token and refresh token later.
---- @return string? state The OAuth state string. This is a sentinel value used to ensure that a request hasn't been
---  forged.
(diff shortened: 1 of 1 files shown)


redgoldlace:
Unfortunately Windows made it _very_ difficult for me to figure out ^^;; looking at it in the Resource Monitor utility made it seem like Firefox was opening multiple connections to the socket in question over the IPv4 loopback adapter, but I'm not certain _why_, nor whether this was even correctly reported; disabling things like HTTPS upgrades or anything else that could be trying to open multiple connections didn't have any effect. I'd thought it might've been related to either the Tailscale or Mullvad clients somehow, but neither of those seemed to affect it. I recall hearing about QT software doing network sweeps a while back and made sure none of it was running, but that made no difference either. At this point I suspect it was probably related to however Tailscale/Mullvad/etc are implemented on Windows - or even some kind of Windows bug downstream of that - but that doesn't explain what would've been responsible in #944 over Wine. As part of trying to figure out a Lua error that wasn't being surfaced properly while debugging this PR, I hooked the SimpleGraphic stuff up to a debugger in Visual Studio and found that it was throwing exceptions in WinRT/KernelBase a fair bit, but considering the rest of the application is.. seemingly fine, I'm not sure what was actually going on.

I was able to confirm that a connection was being opened and immediately dropped by the client though - from my testing, no data was being sent over it, which makes it even stranger.

redgoldlace:
> Sorry about SimpleGraphic not surfacing errors, I just noticed that regression last week and put it on my list to fix/re-enable (would probably start looking in ui_subscript.cpp if you had any remaining curiosity).

I did some digging around yesterday and was able to figure out what the cause of this issue was. For subscripts with a type of `"DOWNLOAD"`, `OnSubError` (in `launch.lua`) will call the subscript callback with `nil, errMsg` as arguments, instead of displaying the error message verbatim. (If the callback returns an error, it then uses *that* to display an error message)

I was able to fix it, but the way I chose to handle cleaning up after errors (the addition of an error handling callback) made things a little messy, so I wound up shelving it. Figured it might save you some time if I mentioned it though ^^

LocalIdentity:
@redgoldlace we have a PoB discord server btw if you'd like to join. Message me (Localidentity) on Discord and I can send you an inv

PR #1309, opened 2025-09-05
Update statMap for Ligthning Conduit

majochem:
Wording for Lightning Conduit shock effect-dependent "more" damage mod was changed with 0.3.0, which caused it to no longer work.

### After screenshot:
image
image


Nightblade:
You might like to fix the spelling in the title, saves Local having to fix it in the changelogs 😁

LocalIdentity (approved):

PR #1328, opened 2025-09-08
fix: update quest rewards for 0.3

Nightblade:
TODO:
1. add area levels for new content.
2. add the penalties for the +5% XP Qimah > The Seven Pillars quest.

Notes:
1. For 2) above, I don't know how to add multi-mod quest reward.
2. The Interludes have all been added as Act 5, and the reward for completing them added as Act 6.
3. The difficulty type needs to be removed.
4. 2) & 3) are probably fine until the quest reward code is rewritten.

### DIFF
--- a/src/Data/QuestRewards.lua
+++ b/src/Data/QuestRewards.lua
@@ -1,7 +1,7 @@
 return {
 	{
 		["Act"] = 1,
-		["Type"] = "Normal",
+		["Type"] = "",
 		["Area"] = "Clearfell",
 		["Stat"] = "+10% to Cold Resistance",
 		["AreaLevel"] = 2,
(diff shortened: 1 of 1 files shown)


Nightblade:
I would like it if the actual quest name was visible.  Perhaps move the area name to a tooltip?

LocalIdentity:
> The following have been moved to comments for now since they require code changes:
> 
> Act 2 > Valley Of The Titans > Medallion > `"+1 Charm Slot"` (both rewards have this) Interlude 2 > Qimah > The Seven Pillars > "5% increased Experience Gain" reward > `"-5% to Elemental Resistances,3% reduced Movement Speed,15% reduced Global Defences,20% reduced Presence Area Of Effect,12% reduced Cooldown Recovery Rate,5% reduced Attributes"`

I added support for these in my latest commit

PR #1355, opened 2025-09-11
Fix subscript errors

Wires77:
This is essentially a re-submit of #1345 that fixes a crash in it.


LocalIdentity (approved):
Doesn't crash me now

PR #1362, opened 2025-09-12
Automatically check if the ModCache needs to be regenerated

Musholic:
Add a github CI job to automatically check if the ModCache needs to be regenerated

You can see an example of the test failing on https://github.com/Musholic/PathOfBuilding-PoE2/pull/2

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,3 +16,19 @@ jobs:
         run: busted --lua=luajit
       - name: Report coverage
         run: cd src; luacov-coveralls --repo-token=${{ secrets.github_token }} -e TestData -e Data -e runtime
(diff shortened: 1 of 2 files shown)


Musholic:
> I think the only issue I have with this is that maybe it should be its own action. We explicitly removed this type of functionality in [PathOfBuildingCommunity/PathOfBuilding#6694](https://github.com/PathOfBuildingCommunity/PathOfBuilding/pull/6694) because it would slow down the tests.
> 
> If you wanted to get fancy, you could check the PR to see if it was updating ModCache already, because if so you wouldn't need to run this check.

The check is actually running in parallel and separately from the run_tests job and does not slow it down, so it's all good on this part

PR #1401, opened 2025-09-16
Guess main socket group based on dps + Full import button

Musholic:
This PR allows to have a one-button full import while allowing to better guess which socket group is the main one based on DPS calculations of all groups. It also includes the main skill in Full DPS.

Note that if we cannot import with only one button, it causes an issue where the variable `mainSkillEmpty = #self.build.skillsTab.socketGroupList == 0` is false because there is a skill coming from the tree ("Called shots" in my case).

This is also useful for the automatic test I'm preparing to have more relevant tests (having a skill correctly included for the full dps stats) (https://github.com/Musholic/PathOfBuilding-PoE2/pull/11).

- [x] I tested the performance on a simple character import (it was very fast), however it needs to be tested on more complex characters before we merge this PR
	- There was some tests done by LocalIdentiy at the time the PR was made with no significant performance impact (around 130ms max delay). There is a possible performance improvements to skip check on non damaging skills but it needs a reliable way to tell if a skill is non damaging.

- Tested with my lvl 35 character

### Before screenshot:
image

### After screenshot:
image

### DIFF
--- a/src/Classes/ImportTab.lua
+++ b/src/Classes/ImportTab.lua
@@ -925,7 +925,11 @@ function ImportTabClass:ImportItemsAndSkills(charData)
 		end)
 	end
 	if mainSkillEmpty then
-		self.build.mainSocketGroup = self:GuessMainSocketGroup()
+		local mainSocketGroup = self:GuessMainSocketGroup()
+		self.build.mainSocketGroup = mainSocketGroup
+		if mainSocketGroup ~= nil then
+			self.build.skillsTab.socketGroupList[mainSocketGroup].includeInFullDPS = true
+		end
 	end
 	self.build.itemsTab:PopulateSlots()
 	self.build.itemsTab:AddUndoState()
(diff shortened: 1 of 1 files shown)


src/Classes/ImportTab.lua
```diff
@@ -1111,17 +1136,22 @@ function ImportTabClass:ImportSocketedItems(item, socketedItems, slotName)
...
+        socketGroup.includeInFullDPS = true
+        local mainOutput = self.build.calcsTab.calcs.buildOutput(self.build, "MAIN").player.output
+        socketGroup.includeInFullDPS = false
+        local dps = mainOutput.FullDPS
+        if dps > bestDps then
+            bestDps = dps
+            bestSocketGroup = i
+        end
+    end
+    return bestSocketGroup
```
Nightblade, line 1154: (these should be tab-indented)

Nightblade: The "Refined GitHub" extension for FF can show whitespace in PRs, and I try to take a quick peek of most PRs, especially interesting ones.  😃

Musholic:
The PR is rebased and ready to be merged IMO, I updated the description regarding the performance tests

PR #1420, opened 2025-09-19
DPI Scaling Added 

LeonSparta:
# Added DPI Scaling Support
Added DPI Scaling Support, tested through scaling factor from 1.0 to 3.0 with a step of 0.25.
- Set a scaling factor through windows
- Swap the dll, main.lua and launch.lua
- Crystal clear texts and UI
### Before screenshot:
image
image
image
image
image
image

### After screenshot:
image
image
image
image
image
image


LeonSparta:
> Finally got around to testing this, found this bug with cursor positioning: image
> 
> This is at 250% scaling, but seems to occur at other scalings as well.
> 
> Moving the cursor makes it worse: image

Sorry for the mess, I'm gonna try to squash this later after I get off work today.

LeonSparta:
Checked the code, same to be related with the l_SetCursorPos and l_GetCursorPos in the simplegraphics dll. Now I know what it does, should be a easy fix, before I was just speculating the adjustment needed for the scaling.

Wires77 (approved):
Turns out there's an existing bug on the released version that messes with the cursor placement at high zoom levels on the Notes tab.  E.g. Ctrl+Scroll up and you'll eventually see something like this:

image

Since it's pre-existing I'm going to merge these PRs, but if you felt like looking into it more I'd be grateful.

PR #1437, opened 2025-09-22
Add support gem filtering options in Skills tab

Blitz54:
Needs more work, at the moment the dropdown does filter by Lineage supports. But the `Highest Tier Per Family` is not implemented yet. Some Gem Families have a Lineage, others do not, and the SupportTier in the dat file shows 0 for Lineage like Brutus' Brain, so we can't just use the highest number. We also can't look for 3, as some families only have 2. Need to do some wacky combination probably.

Also the isLineage is true for a bunch of persistent support gems. So slight change to exporter so that it looks for the flavour text too, as only true lineage have flavour text.

### After screenshot:
image


Blitz54:
Changes added to #1743

PR #1473, opened 2025-10-02
Fix Fanatic Bow, Bolting Quarterstaff missing flat Elemental Damage

LocalIdentity:
We were not exporting the hidden implicit mods for flat added elemental damage
We were also not exporting the spear throw line for all spear bases
While going through the staves and shields, I realised a lot of the bases were being duplicated so I fixed that too


src/Export/Scripts/bases.lua
```diff
@@ -132,6 +132,9 @@ directiveTable.base = function(state, args, out)
 			table.insert(implicitLines, line)
 			table.insert(implicitModTypes, modDesc.modTags)
 		end
+		if mod.Id == "SpearImplicitDisplaySpearThrow1" then
+			table.insert(implicitLines, "Grants Skill: Spear Throw")
```
Wires77, line 136: Interesting, so spear throw isn't an ItemInherentSkill like the others?

LocalIdentity: For some reason, it is not. Maybe because it doesn't have more than 1 level?

PR #1489, opened 2025-10-07
Add new feature for automatic attribute allocation

### LINKED ISSUE #1690: Add a Button to Optimize Travel Node Attributes
### Check for duplicates

Windows

It is time consuming to redo travel node attributes each time gear and skills change.

Pressing a button that reselects the selectable attributes of travel nodes in a way that matches the requirements of your currently equipped gear and skills.

eg: Build is made selecting only STR nodes.
Build requires 150 dex, 50 str and 25 int. Gear grants 50 dex.
Pressing the button will start reselecting attribute nodes.
It will first try to select attributes to enable equipped gear.
It will then select DEX till the attribute deficit is 50 dex. 
Then it will alternate between DEX and STR in till the deficit is 25.
It will then evenly split attributes till out of travel nodes.

If attribute requirements are met, remaining travel nodes will be equally split between all attributes. ( Possible setting adjustment )

If attribute requirements are not met and there is no remaining pathing nodes, the process will complete but throw a warning.

A way to change pathing attributes in bulk without clicking every one.

A checkbox that dynamically performs this action each gear/skill change. 

### Additional context

Possible issues with meeting attribute requirements of gear that grant attributes.
Possible issues with pathfinder's "Traveler's Wisdom"
Possible issues with builds that was to stack one attribute. 
Possible issues with weapon swap/weapon swap trees/Gear with attributes on weapon swap.

majochem:
I was getting a bit annoyed by having to micro-manage attributes when planning out passive trees, so I added the option for some automation.
 
Until now assignment of attributes for travel nodes worked like this:
1. You select attribute via pop-up
2. You hold one of three hotkeys to force an attribute
3. If you allocate more than one node, it allocates all attribute nodes in the path to your last selected attribute

I found that inconvenient because I would have to keep track of how many attributes I have vs. how many I needed. Especially on triple attribute builds, it felt pretty tedious.

### Feature Description
Now there is a new menu button called "Auto Attribute Config" *(I'm open to better naming suggestions)* on the bottom of the tree tab and it looks like this:

*Button to open Config:*
image

*Actual Config dialog:*
image

*Confirmation dialog when re-allocating all:*
image

#### Quick summary explanation:
- Allows setting target weights and max values for attributes that are used for travel attribute passives
- Configs can be set to be weapon set specific
- Weights can be auto calculated from attribute requirements
- There is an option to "re-allocate all", which re-allocates all passives for the current set

**Full copied explanation from `help.txt`:**
> You can enable the automatic allocation of attributes via the "Auto Attribute Config" button at the bottom
of the "Tree" menu section. Each configuration is saved per tree and per weapon set. So if you have multiple
trees, each will have its own weapon set configuration values.
> 
> *Weights:*
If enabled, attribute travel nodes will automatically be assigned to Strength / Dexterity / Intelligence
according to your configured "weight" values. E.g. values of Str: 1 / Dex: 1 / Int: 2, would result in
roughly 25% of the attribute nodes being assigned to Strength and Dexterity and 50% to Intelligence.
By default, attributes gained from items and other small passive nodes are taken into account when
calculating the actual vs. desired attribute ratios.
> 
> *Max Value:*
If a "Max Value" is entered and the "Limit to Max?" checkbox is ticked, no more nodes will be allocated
to that attribute, once the maximum value is reached
> 
> *Attribute Requirements:*
For ease of use, the "Use Attribute Requirements" checkbox can be ticked. This will result in weights
automatically being based on current attribute requirements from gems and gear.
> 
> *Item Mods:*
If you want your attribute allocation to be gear-agnostic, you can tick the "Ignore Attribute Requirements"
checkbox. Any attribute bonuses gained from equipment will then not be taken into account during allocation.
Note: This does not affect modifiers to attribute requirements found on gear.
> 
> *Use for both Weapon Sets:*
If you want the settings from one weapon set to be used for both, you can tick the "Use for both Weapon
Sets" checkbox. This will use the settings from the current set regardless of active weapon set. 
Note: Attribute requirements will still be based on each respective weapon set.

### Other Functionality:
- Saved Auto Attribute Config values are written to and retrieved from the XML
- Added hint about new automatic allocation to `ModifyAttributePopup`
image

- Added hint about active automatic attribute allocation when hovering over attribute nodes (in case user is loading someone else's build and confused what's happening)
image

- Added information about Auto Attribute Config to the `help.txt`
- Added tab navigation to the Auto Attribute Config pop-up to easily switch between attributes

**Test cases**

- [x] Basic Popup
	- [x] Opens and closes correctly
	- [x] Works in different window ratios
	- [x] option controls only enabled if main `enabled` box is checked
- [x] Item mods
	- [x] Correctly accounted for (`BASE`, `INC`, and `MORE`)
	- [x] Ignored if desired
- [x] Use Attribute Requirements
	- [x] Calculated correctly
	- [x] Absolute values are fulfilled before considering relative ratios
	- [x] Updated when requirements change, both for calculation and in popup display
	- [x] Modification of weight values disabled, if using requirements
- [x] Special Jewels
	- [x] Timeless jewels
		- *No mods affecting attributes afaik?*
	- [x] Time-lost jewels
		- [x] %increased effect of notables/small passives is taken into account for attribute totals
	- [x] Intuitive leap likes
		- [x] From Nothing
		- [x] Controlled Metamorphosis
		- [x] Auto allocation works
		- [x] Manual allocation (hotkey or right click swap works)
	- [x] Against The Darkness
		- [x] %attribute mods correctly accounted for
- [x] Alt Pathing (via holding shift)
	- [x] auto allocation works
	- [x] non-attribute-passive mods are correctly accounted for
- [x] Saving Config
	- [x] One Tree
	- [x] Multiple Trees
	- [x] Enable "Save" on config change
- [x] Loading Config
	- [x] One Tree
	- [x] Multiple Trees
	- [x] No existing config
	- [x] Works if `calcsTab.mainOutput` doesn't exist yet
	- [x] Works if tree gets converted to different version
- [x] Weapon Swap passives 
  - [x] Allocation *mostly* works. Weapon-set specific passives will respect the configured ratios and disregard stats from the tree for the currently non-active weapon set (b/c those don't affect actual attribute totals unless that weapon set is active).
  - [x] Re-allocating all nodes only affects the generic passives and the currently active weapon passives
- [x] Hotkeys
	- [x] "1/2/3" & "S/D/I" still works
	- [x] Right click attribute swap still works
- [x] Undo States work
- [x] Copy Tree --> copies configs as well
- [x] Reset Tree Behavior
	- [x] Result: Config does not get reset (intended)
- [x] Import Behavior
  - Importing from build code with config, correctly imports config
  - Importing via build code without config sets it to nothing (overwriting if one exists, but I think that's fine?)
  - Importing from account keeps current config unchanged

### Limitations:
- Automatic allocation for weapon-set specific passives is only active, if that weapon set is currently also active. That is because attribute values and requirements are not currently easily accessible for a non-active weapon set. So 
    - Allocating generic passives will work
    - Allocating weapon set 1 passives, while weapon set 1 is active will work
    - Allocating weapon set 2 passives, while weapon set 1 is active will fall back to original manual allocation logic
- I have some ideas for solving this in the future, but this PR is already quite big, so I think it should be solved separately first

[Test Build with some relevant test items](https://pobb.in/liwJvh6E-Wqc)


majochem:
Back to draft because need to fix config getting lost on tree version conversion and also want to make sure it works with #1665

majochem:
Marking as ready for review again:
- fixed a bunch of stuff
- enabled weapon-set specific configs
- improved compatibility with non-standard values on travel passives (e.g. PF Traveler's Wisdom). You cannot set weights for those (too much of an edge case), but they shouldn't cause issues and they get actively ignored when using "re-allocate all"

PR #1544, opened 2025-12-09
added jewel ranges for passive tree nodes while holding ALT (Oracle QoL)

Gisbert12843:
added the ability to hold alt while hovering a passive tree node to show jewel distance circles and highlighting nodes in their respective ranges

(to accommodate for the upcoming "Traveler's Wisdom Ascendancy Node of the Oracle)

(The Screenshot also holds "Ctrl" to hide the node info)

Github Diff sadly differs from the VSCode one. Comparing my changes to origin shows that i only inserted lines at 2 points in the file and did not formatted/adjusted others. Githubs Diff appears to be misleading here.

image

### DIFF
--- a/src/Classes/PassiveTreeView.lua
+++ b/src/Classes/PassiveTreeView.lua
@@ -168,11 +168,11 @@ function PassiveTreeViewClass:Draw(build, viewPort, inputEvents)
 	local offsetY = self.zoomY + viewPort.y + viewPort.height/2
 	local function treeToScreen(x, y)
 		return x * scale + offsetX,
-		       y * scale + offsetY
+			   y * scale + offsetY
 	end
 	local function screenToTree(x, y)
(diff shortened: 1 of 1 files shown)


Nightblade:
Please set your editor to TAB indents for this code-base.

Gisbert12843:
If you need me to address this weird spell check pipeline tell me please, would really like my feature to be implemented in an upcoming beta as soon as you find time

Wires77:
A couple comments after testing it out:

- Probably only need it to work when hovering keystones right now.  If there are other changes in the future it'd be easy enough to change that conditional again
- In a similar vein, just show the medium radius when hovering keystones.  Showing them all is likely to confuse just as many users as it helps
- Any reason we didn't just use the existing shading seen when hovering jewel sockets?  I'd like to be consistent here

PR #1613, opened 2025-12-15
Added sources of gaining rage

MrHB212:
Screenshot 2025-12-16 004331
Screenshot 2025-12-16 004357


src/Modules/ModParser.lua
```diff
@@ -2707,11 +2707,16 @@ local specialModList = {
 	["gain %d+ rage on hit with axes or swords"] = {
 		flag("Condition:CanGainRage"),
 	},
-	["gain %d+ rage on melee hit"] = {
-		flag("Condition:CanGainRage"),
-	},
+	["gain (%d+) rage on melee hit"] = function(num) return {
+			mod("RageOnMeleeHit", "BASE", num),
```
LocalIdentity: Should use ModFlag.Melee as a way to only have it work on Melee skills

src/Modules/CalcSections.lua
```diff
@@ -1803,6 +1803,8 @@ return {
 	{ label = "Total", { format = "{0:output:Rage}", }, },
 	{ label = "Rage Effect", { format = "{1:output:RageEffect}", { modName = "RageEffect" }, }, },
 	{ label = "Maximum Rage", { format = "{0:output:MaximumRage}", { modName = "MaximumRage" }, }, },
+	{ label = "Gain on Melee hit", { format = "{0:output:RageOnMeleeHit}", { modName = "RageOnMeleeHit" }, }, },
```
LocalIdentity: Should change to be Gain on Hit so it can work for other sources

src/Modules/ModParser.lua
```diff
@@ -2707,11 +2707,16 @@ local specialModList = {
...
+			flag("Condition:CanGainRage")
+	}end,
+	["gain (%d+) rage when hit by an enemy"] = function(num) return {
+			mod("RageWhenHit", "BASE", num),
+			flag("Condition:CanGainRage")
+	}end,
 	["gain %d+ rage on melee weapon hit"] = {
-		flag("Condition:CanGainRage"),
+		flag("Condition:CanGainRage")
```
LocalIdentity: Can use the weapon modflag for this similar to how we handle other weapon hit mods

LocalIdentity (changes requested):
Just needs to change to be more generic so in the future we can support other  mods like range on spell hit or range on arrow hit etcc

src/Modules/CalcSections.lua
MrHB212: done

src/Modules/ModParser.lua
MrHB212: done

src/Modules/ModParser.lua
MrHB212: I actually deleted that parse since it seemed like a duplicate to me, the nodes still work fine without that.

MrHB212:
image

Will also look into the Rage supports now since they dont actually do anything.

Edit: One of the nodes show "Rage on Hit" instead of "Rage on Melee Hit", dont know why.

MrHB212:
### Rage Support + Maul + Fury of the Mountain
image
image
image
image
image

MrHB212:
I tested some custom mods with different ModFlags and it seems to work fine:

Gain 5 Rage on Spell hit
image
--

Gain 5 Rage on Slam Hit + on Melee Hit nodes
image
--

Gain 5 Rage on Slam Hit + Fire Hit (Fire Attunement Support)
image

PR #1622, opened 2025-12-16
Add support for removing all of a file type from manifest

LocalIdentity:
Adds support for removing all of a file type from the manifest script
Recent update accidently included some csm files in the manifest which broke the update
We never care about the csm files so fine to always remove them


Wires77 (approved):
Probably could've still kept the functional style, just augmenting _exclude_file to take in a glob format

PR #1625, opened 2025-12-16
Speedup export of skill tree and regenerating files

LocalIdentity:
4x speed improvement on passive tree generation and 2x speed improvement on regenerating files

We now pass a manifest file to the ooz extractor so we are no longer limited by the cmd character limit and can batch many previous calls into a single call
This cuts our regen of files from 7 cmd calls taking 30s to 2 cmd calls taking 15s
For the passive tree it cuts it from 18 cmd calls taking 45s to 2 cmd calls taking 11s

Also removed the old logic for non dat64 files as we don't need it any more

In the passive tree file I batched together all the asset extraction calls and removed a whole bunch of prints to the console so now the console isn't flooded

### DIFF
--- a/src/Export/Classes/GGPKData.lua
+++ b/src/Export/Classes/GGPKData.lua
@@ -44,12 +44,8 @@ local GGPKClass = newClass("GGPKData", function(self, path, datPath, reExport)
 	self.dat = { }
 	self.txt = { }
 	self.ot = { }
-
-	if USE_DAT64 then
-		self:AddDat64Files()
-	else
-		self:AddDatFiles()
-	end
+	
+	self:AddDat64Files()
 end)
 
 function GGPKClass:CleanDir(reExport)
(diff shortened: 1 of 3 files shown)


src/Export/Classes/GGPKData.lua
```diff
@@ -67,48 +63,44 @@ function GGPKClass:ExtractFilesWithBun(fileListStr, useRegex)
 	os.execute(cmd)
 end
 
+-- Use manifest files to avoid command line limit and reduce cmd calls
+function GGPKClass:ExtractFilesWithBunFromTable(fileTable, useRegex)
+	local useRegex = useRegex or false
+	local manifest = self.oozPath .. "extract_list.txt"
```
Wires77, line 69: Where does this file come from?

LocalIdentity: This is just the path to the file. The next line creates it as a temp file before removing it at the end

Wires77: Oh sure, I misunderstood what you were trying to do here, makes sense.

Wires77 (approved):

PR #1645, opened 2025-12-18
Trader Improvements: Pin FullDPS, HitPool and Rarity to the Top, Add rarity comparison, Show only target Ring

falleng0d:

The aim of these changes is to make the trade reflect the real world usage of players by: 
- Pinning the important stats to the top of the comparison (when you are checking a lot of items in the listing, the stats constantly shift up and down depending on the values, this makes it very slow to find what you want, because you're constantly losing track of the actually important stats, which are usually either DPS, defenses, rarity or a mix or these. by pinning the stats to the top, you can keep your eyes focused on the same area as you go down the listing and immediately see what is important)
- Adding a new stat to the diff (Item Rarity) 
- Show only the target ring when just playing comparisons for ring stats
- Display the weights at the end of the diff, this is extremely useful for tuning the importance of the desired search weights (we could optionally make this visible only when holding shift or alt, for me i found out that it's always useful to see them, but just let me know)

just run runtime\Path{space}of{space}Building-PoE2.exe

### Before screenshot:
image

### After screenshot:
image

This is my first time sending a merge request here and i don't know what are the expectations and if you guys even want any of these changes merged. i made these mostly for me and my usage patterns, but if you find only a part of them are useful or even none, just let me know and i can adjust. I am sending this mostly to give it back to this amazing community but i won't make a big deal out of it, Thanks!


vaisest:
> i don't know what are the expectations and if you guys even want any of these changes merged

I definitely think the score tooltip idea is good. There have been a couple scenarios where I've not even realised that my weights are so wrong that the weights don't count dps at all. Though it took me a couple minutes to realise it's behind pressing shift. Dunno if they could just be next to the stat comparisons themselves or something?

Pinning the stats also seems good. Stat comparisons are pretty hard to read

PR #1648, opened 2025-12-18
Added Hinder and Maim debuff in config and the immune node

MrHB212:
immunenodes
config


LocalIdentity (changes requested):
Maim and Hinder need to be implemented as debuffs in CalcPerform with the rest of the buffs / debuffs

PR #1660, opened 2025-12-19
Fix Artillery Ballista missing skill part name

mateokladaric:
Fixes #noissue.

Currently Artillery Ballista adds a no name Damage Source
image
Empty field under Artillery Option

image

Full DPS damage breakdown with the empty name for source.

- Built and loaded the same build code

https://pobb.in/QiZiGiYLs5ID

### Before screenshot:
image

### After screenshot:
image


LocalIdentity (changes requested):
This most likely requires changes to the export scripts for skills to grab the relevant labels for certain scenarios

labels are not set by hand as all the data should be in the game files, we just need to export it properly

PR #1665, opened 2025-12-19
Add Traveler's Wisdom Support

justjuangui:
### Description
This PR:
- Support "dyamic" attributes
- Improve "rotation" attrribute with right click
- Support for "unlockBy" in options for attributes nodes
- Check for invalidate attribute selection
- Backward Compatibility for previous save
- Import & Save support

### Format with backward compatibility
image

### Evidence

image

image

image


Nightblade:
Please fix spelling before this is merged.  ❤️

tomatosource (commented):
Fix spelling mistakes

FinalLightNL:
was about to request this as a update, glad someone is working on it.
thanks POB team 🙏

tomatosource:
@Nightblade / @justjuangui Hey guys - HNY :) 
Any chance of getting this merged? 

I fixed  up spelling mistakes on a branch here if that makes life easier: https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/1682

justjuangui:
> @Nightblade / @justjuangui Hey guys - HNY :) Any chance of getting this merged?
> 
> I fixed up spelling mistakes on a branch here if that makes life easier: #1682

HNY too,

thx for the help, but like Local said we was on vacation, and i just arrive, but just to keep in mind, even if i fix, it doenst mean that it will merge soon, because depends on maintainers :D, this week i will fix this PR in my branch.

Wires77 (changes requested):
Fixed the spelling for you.  Do we know what happens in game when you deallocate this node?  Right now those nodes turn into "+5 to any attribute", which isn't ideal.

Looks like this also broke the hotkeys specified in the popup menu
image

Finally, we should parse the contents of the ascendancy node so people know it's implemented

majochem:
image

Seems like the reset to "any attribute" is till happening when de-allocating the Ascendancy node and it doesn't convert back when you re-allocate the Ascendancy node either (although it works with undo). In-game it doesn't let you de-allocate the Ascendancy node if you have any converted nodes, so they don't have to deal with the problem, but I don't think that's a good option for PoB.

Should be relatively easy to fix, if we're fine with a generic fallback logic like "always strength", "always random attribute" or "damage -> str | defences -> dex | cost -> int". It's a bit more complicated if you want something fancy. In theory you could keep track on the node what it was converted from, but not sure if that's worth the extra effort.

PR #1669, opened 2025-12-20
Added ascendancy nodes: No Mercy + Corrupted Lifeforce

MrHB212:
image
image
image
image


Wires77:
These checkboxes could probably be more useful.  Blood Boils applies to nearby enemies always that aren't unique, so we could just assume the enemy has them unless "Is the enemy a boss?" is true.

Concentration is harder, because I think people would want to see on average how much damage it adds when determining whether to allocate the node.  A lot of bosses have immunity phases where they could regenerate the Concentration, but I still think an average based on 30% of their life makes more sense.  It could be similar to culling strike.

MrHB212:
image
image
image


> Concentration is harder, because I think people would want to see on average how much damage it adds when determining whether to allocate the node. A lot of bosses have immunity phases where they could regenerate the Concentration, but I still think an average based on 30% of their life makes more sense. It could be similar to culling strike.

Changed the config to be a count instead with a default value of 50. Im not sure but i guess it scales linearly.

PR #1680, opened 2025-12-26
Add parsing for 'Forced Outcome' oracle passive, minion presence passives, and bugfix for Inspiring Ally

Dimencia:
- Adds support for 'Forced Outcome' oracle passive (Inevitable Critical Hits)
- Adds support for 'Pack Encouragement' and 'Licking Wounds' passives (or anything that scales with number of minions in your Presence)
- Fixes issue where "increased life and mana regeneration rate" would give +life, and nothing for mana, instead of regen (but it's only on one node anyway)
- Fixes longstanding bug with 'Inspiring Ally' incorrectly adding damage from non-companion minion nodes
- Updates companionInPresence config condition to only appear when you have nodes that need it, instead of whenever you have a companion

Just wanted to get those passives working so I could plan my build, but the changes are probably straightforward enough to merge in for everyone

### DIFF
--- a/src/Data/ModCache.lua
+++ b/src/Data/ModCache.lua
@@ -2784,7 +2784,7 @@ c["5% increased Area Damage"]={{[1]={flags=512,keywordFlags=0,name="Damage",type
 c["5% increased Area of Effect"]={{[1]={flags=0,keywordFlags=0,name="AreaOfEffect",type="INC",value=5}},nil}
 c["5% increased Attack Critical Hit Chance per 10 Tribute"]={{[1]={[1]={actor="parent",div=10,stat="Tribute",type="PerStat"},flags=1,keywordFlags=0,name="CritChance",type="INC",value=5}},nil}
 c["5% increased Attack Damage"]={{[1]={flags=1,keywordFlags=0,name="Damage",type="INC",value=5}},nil}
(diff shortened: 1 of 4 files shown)


Dimencia:
Hold on this, the Forced Outcome calcs are wrong (using INC instead of MORE, and the skill clearly says 'less'), and I want to double check a few things as I make that change, because it seems odd that having it is worse than not having it at <25% crit chance or so

majochem:
You're using average number of re-rolls, which is not the correct approach. In your case, any expected value over 3.3 will give you 0% effective crit multi, which is not true. Even if you just have 5% crit chance, you'll still get 100% of your crit bonus in 5% of cases, so you know it can't be true.

Afaik, you only care about the chances for different scenarios

1. "no re-roll", i.e. crit on first hit, which is just your default effective crit chance
2. 1 re-rolls, -> 30% less multi
3. 2 re-rolls -> 60% less multi
4. 3 re-rolls -> 90% less multi
5. Anything else is 100% less

And then you calculate the average expected crit multi from the proportions of those chances.
(At least that is how I understand it)

Maybe you can join the Discord (if you haven't already) though. There were still some open questions regarding the interaction with "Bifurcated" critical hits iirc.

LocalIdentity:
Also @Dimencia the issues and additions you made you be spread into 4 separate PRs with the 2 companion ones being together and the other ones being on their own

PR #1686, opened 2025-12-28
Fixed names for several ascendency root nodes

Nickomang:
For several ascendencies, the root node has an outdated name. This PR fixes the following:
- "Necromancer" -> "Lich"
- "Gambler" -> "Gemling Legionnaire"
- "Master of the Elements" -> "Invoker"
- "Brute" -> "Warbringer"

### Before screenshot:
before

### After screenshot:
after


LocalIdentity (changes requested):
The tree file is directly generated from the game files so you need to make the relevant change to the export script so your manual change does not get overridden https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/dev/src/Export/Scripts/passivetree.lua

It’s fine to manually edit the 0.1, 0.2 and 0.3 tree files though

Wires77 (approved):

PR #1694, opened 2025-12-30
Add support for Arakaali's Lust

ain9156:
Adds support for Arakaali's Lust hit damage scaling per Poison stack on the enemy

- Added a skill that hits
- Added Poison Support
- Added Arakaali's Lust
- Adjusted the number of Poison stacks on the enemy

https://pobb.in/9Qsv16yyc9V2

### Before screenshot:
before2n

### After screenshot:
after2

### DIFF
--- a/src/Data/Skills/sup_dex.lua
+++ b/src/Data/Skills/sup_dex.lua
@@ -410,6 +410,14 @@ skills["SupportArakaalisLustPlayer"] = {
 			label = "Arakaali's Lust",
 			incrementalEffectiveness = 0.054999999701977,
 			statDescriptionScope = "gem_stat_descriptions",
+			statMap = {
+				["support_debilitate_hit_damage_+%_final_per_poison_stack"] = {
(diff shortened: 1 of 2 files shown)


src/Data/Skills/sup_dex.lua
```diff
@@ -410,6 +410,14 @@ skills["SupportArakaalisLustPlayer"] = {
 			label = "Arakaali's Lust",
 			incrementalEffectiveness = 0.054999999701977,
 			statDescriptionScope = "gem_stat_descriptions",
+			statMap = {
+				["support_debilitate_hit_damage_+%_final_per_poison_stack"] = {
+					mod("Damage", "MORE", nil, 0, 0, { type = "Multiplier", actor = "enemy", var = "PoisonStacks", limitVar = "ArakaaliMaxHitDamage", limitTotal = true } ),
```
LocalIdentity: Needs to have ModFlag.Hit in the place of the first 0 after the nil so that the damage only scales hits

LocalIdentity (changes requested):
Just needs the flag so it only applies to hits

Wires77 (approved):

Issue #1441, opened 2025-09-23
Inspiring Ally incorrectly applies to Minion nodes (e.g., Power of the Dead, Emboldening Lead)


Sionius1406:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

Inspiring Ally says: “Increases and Reductions to Companion Damage also apply to you.” It should only affect Companion modifiers, not Minion modifiers. Currently, Path of Building applies Inspiring Ally to generic Minion nodes like Power of the Dead and Emboldening Lead, which is incorrect.

**Steps to reproduce**

1. Allocate a Minion node (e.g., Power of the Dead or Emboldening Lead).
2. Allocate Inspiring Ally.
3. Observe player damage increases as if Minion increases were Companion increases.

**Expected**
Only Companion Damage modifiers are mirrored to the player.

**Actual**
Minion nodes are treated as eligible.

### What is the value from the calculation in Path of Building?

**Steps to reproduce**

1. Allocate a Minion node (e.g., Power of the Dead or Emboldening Lead).
2. Allocate Inspiring Ally.
3. Observe player damage increases as if Minion increases were Companion increases.

### How to reproduce the issue

**Steps to reproduce**

1. Allocate a Minion node (e.g., Power of the Dead or Emboldening Lead).
2. Allocate Inspiring Ally.
3. Observe player damage increases as if Minion increases were Companion increases.

### PoB for PoE2 build code

```shell
[...]

PR #1697, opened 2025-12-31
Fix Inspiring Ally considering all minion damage increases, instead of only companions

Dimencia:
Fixes longstanding bug: #1441 

Tags companion skill mods with the CreatesCompanion SkillType, so we can filter to only those skills in CalcOffence for CompanionDamageAppliesToPlayer

Issues:

1. Only the MinionModifier LIST mod for companions is tagged with SkillType CreatesCompanion, not the individual mods
2. ModStore:List does not consider SkillType of the LIST mods it's processing
3. In EvalMod, SkillType is only used as a filter if the mod has one specified.  If it does not, it will match even if cfg has SkillTypes specified

I considered adjusting the behavior of :List and/or EvalMod to filter on a specified SkillType, but was not sure if that functionality is intended or what side effects it might have.  The simplest and safest fix seems to be adding the SkillType tag to each mod in the MinionModifier list (for companion nodes), and filtering it inside of CalcOffence

I'm not completely sure of how this might affect those companion node mods, because I'm not completely familiar with how minion mods work - but it would make sense for their effects to only occur if the player has a CreatesCompanion skill


LocalIdentity:
Superceded by https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/1884

PR #1704, opened 2026-01-03
Fix low life defence mods not applying sometimes when using Starkonja's, Rise of the Pheonix and Coward Legacy

NamNgHH:
Fixes # 1474.

The issue was to fix the Coward's Legacy belt lowlife modifiers not working correctly. This was wrong since it worked correctly, verified by using the Pain Atunement note that 75% Lowlife is accounted correctly with Beidat's Hand. So the thing that doesn't work is the defense's calculation of items like Starkonja's Head or Rise of the Phoenix. So I solved the problem by changing calcs.defence functions. 
- Before the changes, the build is equipped with Starkonja's Head, Rise of Phoenix, and the evasion rating and fire resistance don't change when Beidat's Hand is taken, and having Coward's Legacy equipped. 
- Currently, if I take out Coward's Legacy, the lowlife effect of Starkonja's and Rise of the Phoenix doesn't affect my character
- If I unallocate Beidat's Hand, the lowlife effect will also stop working since you are not a lowlife anymore. 

https://pobb.in/R8tUNGCcbxCR
### Before screenshot:
image
image

### After screenshot:
image
image

### DIFF
--- a/src/Modules/CalcDefence.lua
+++ b/src/Modules/CalcDefence.lua
@@ -951,9 +951,9 @@ function calcs.defence(env, actor)
 		breakdown.Mana = { slots = { } }
 		breakdown.Spirit = { slots = { } }
 	end
-	if actor == env.minion then
-		calcs.doActorLifeManaSpirit(env.minion)
-		calcs.doActorLifeManaSpiritReservation(env.minion)
+	if actor == env.minion or actor == env.player then
+		calcs.doActorLifeManaSpirit(actor)
+		calcs.doActorLifeManaSpiritReservation(actor)
 	end
 
 	-- Block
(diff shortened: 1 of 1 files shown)


Wires77 (commented):
I think the proper fix looks something like this: https://github.com/PathOfBuildingCommunity/PathOfBuilding/pull/9351

That PR is for PoB1, and all this code was changed in https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/79, but we may just want to calculate these values earlier totally.  This may allow us to replace the lower calls to doActorLifeManaSpirit.  Doesn't seem to make much sense to call it so many times.

I'm trying to keep parity with the ordering of calculations, since we've divined them slowly over the years and moving things around in the Calc files causes frequent bugs if we get the ordering wrong.  I think the way you have it in this PR would break any calculation like https://www.poewiki.net/wiki/Crimson_Power if they added it to PoE2

NamNgHH:
The mentioned method in the PoB1 issue fixed the bug, so I reverted my original changes and put in the same fix.

LocalIdentity (approved):

PR #1707, opened 2026-01-05
Add support for Rageforged support gems

laFleuere2:
Add support for Rageforged support gems
- Add a skill with Rageforged support;
- Add "Gain 1 Rage on Hit" Custom modifier (or add additional skill with rage support);
- Set Rage >= 10 in Configuration tab;

### Before screenshot:
image

### After screenshot:
image


LocalIdentity (changes requested):
The support grants generic more damage not hit damage so you need to remove the modflag

laFleuere2:
> The support grants generic more damage not hit damage so you need to remove the modflag

fixed.

Wires77 (approved):

PR #1716, opened 2026-01-11
Add 0.4 Uniques to item DB

### LINKED ISSUE #1687: no unique talismans
### Check for duplicates

Windows

You have have the rare talismans but there are no unique talisman have been added to pob poe 2

add unique talisman to pob poe 2

_No response_

### Additional context

_No response_

Blitz54:
Fixes #1687

EDIT: The uniques have been added in a separate PR. This likely should not be merged, a new PR with the unique support should be created.

Adds the 11 new uniques to the DB. Some custom mods still need supporting. And Atziri/Vaal Cultivated items do not have the header icons currently as we need to find a way to do that nicely.

Also add support for Darkness Enthroned runes, (not the effect%), Bringer of Rain "as if a body armour"

Also Atziri's Splendour only soul cores mod.


Blitz54:
Only thing holding this back is the soul core and socket scaling from darkness enthroned and Mahu shield. Currently any runes that affect minions or enemies cause an error when selected in the drop down. And pasting items does not properly find the runes, which was always the case. A few issues with it there. Eg, 8 strength rune scaled to 16 with darkness enthroned will have the rune matcher think it's two 8 strength runes, instead of one scaled rune alongside another rune. It just ignores the other rune lines.

Also the bonded mods that provide life/mana do not scale. But bonded mods like elemental threshold do. I have no damn clue why.

Another issue is bonded mods combine with the default rune. The 35 mana rune breaks on darkness, as it can't find a match when it becomes 45 mana. But the life rune doesn't have the same issue somehow. Bonded mods should stay separate regardless.

PR #1725, opened 2026-01-15
Add shapeshift condition to ConfigOptions

Burundai:
Conditional:Shapeshifted is not always making it downstream to calculations. Added a bypass configuration option "conditionShapeshifted" to force the shapeshifted conditional to true. Tested locally with various nodes and corroborated toggle effect on calculations sheet. Can provide screenshots if needed!


Dimencia:
Can you give examples of some scenarios where it doesn't make it to calculations?

An override toggle seems like a decent idea to me, but if something isn't setting the condition when it should, then we should probably fix that too

Blitz54:
Superseded by #1778. Thanks for the PR though, guess it got lost over time.

PR #1778, opened 2026-03-06
Fixes unsupported Thorns calculations and related support interactions

DS-Koala:
Adds support for thorns as a standalone damage source and connects that to relevant support interactions. There is now a "Thorns" skill that will appear if you have thorns damage in your build.

Also added a config checkbox for "Currently Shapeshifted?", since shapeshifting isn't only used for active skills (defensive stats and the "Rough Carapace" node check to see if you're shapeshifted).

Includes support for:
- Typed thorns damage (physical, fire, cold, lightning, chaos)
- % increased Thorns Damage
- % increased Thorns Critical Damage Bonus
- +% / increased Thorns Critical Hit Chance
- Thorns Damage has % chance to ignore Enemy Armour
- Gain Physical Thorns Damage equal to % of Item Armour on Equipped Body Armour
- Gain Physical Thorns Damage equal to % of Maximum Life while Shapeshifted
- Barbs I / II / III
- Quill Burst
- Proper shapeshift condition config support for "while Shapeshifted" effects
- Icetomb Mail Vestements (Gain Cold thorns damage equal to X% of your maximum Mana)
- Crackling Barrier (Gain Lightning thorns damage equal to X% of your maximum Mana)
- Stag Idol's Bonded modifier (Thorns Damage has 40% chance to ignore Enemy Armour)

- Several thorns nodes were unsupported
- Thorns runes were unsupported
- Thorns damage sources were unsupported
- Thorns support gems (Barbs, Quillburst) were unsupported
- Non-skill gem related "while Shapeshifted" mods did not apply.

- Verified flat typed thorns damage parses correctly and appears in Thorns calcs/breakdown
- Verified % increased Thorns Damage, crit chance, crit multiplier, and ignore-armour modifiers affect Thorns correctly
- Verified Barbs I / II apply the thorns payload to supported melee hits and Barbs III applies it twice
- Verified Quill Burst uses the same thorns payload as a triggered AoE skill
- Verified "Gain Physical Thorns damage equal to % of Item Armour on Equipped Body Armour" updates Thorns damage
- Verified "Gain Physical Thorns damage equal to % of maximum Life while Shapeshifted" works with the new shapeshift config condition
- Verified Caltrops only inherits thorns modifiers, not the full thorns payload
- Verified breakdown attribution for added min/max and crit-related thorns sources

https://pobb.in/icDxyhOiMSLm

### After screenshot:
ThornsSkill
SupportDPS
ThornsSourceCalcs
ShapeshiftedConfig
NodeSupport


DS-Koala:
Made a small tweak to future-proof the Thorns skill; it will now be displayed if you have any source of "PhysicalThornsMin", "ChaosThornsMax", etc.

src/Data/Skills/other.lua
```diff
@@ -228,6 +228,86 @@ skills["AcidicConcoctionPlayer"] = {
...
+	statSets = {
+		[1] = {
+			label = "Thorns",
+			incrementalEffectiveness = 0,
+			statDescriptionScope = "skill_stat_descriptions",
+			baseFlags = {
+				thorns = true,
+			},
+			baseMods = {
+				flag("CannotBleed"),
```
krryan: I believe this is incorrect (though there is no way to demonstrate it available in-game right now). Rather, the issue is that the player’s modifiers that increase the chance of an ailment, or that guarantee that you inflict an ailment, don’t apply to Thorns. I would have to double-check (been a while since I tested it) but I believe effects changing ailment magnitudes are also ignored.

So chill, freeze, flammability, ignite, and shock can happen because cold, fire, and lightning damage have an inherent chance to do those things. Bleed and poison cannot happen because physical and poison damage have no inherent ability to inflict them, and everything that would add them doesn’t apply to Thorns.

I mention this because
1. future-proofing, in case some ability to bleed or poison with Thorns becomes available
2. your calculations for the chances of other ailments are wrong if they’re not ignoring those mods

I think ailment calculations are going to have to be redone for, or have a special case for, Thorns hits. This applies both to the default Retaliate with Thorns, and Barbs or Quill Burst.

src/Modules/ModParser.lua
```diff
@@ -3437,16 +3437,70 @@ local specialModList = {
 	["your skills deal you (%d+)%% of mana cost as (.+) damage"] = function(dmgMult, _, dmgType) return {
 		mod("ScoldsBridleSelfDamage", "LIST", {dmgMult = dmgMult, damageType = dmgType})
 	} end,
-	-- Thorns
+
+	-- Thorns base damage
+	["(%d+) to (%d+) physical thorns damage"] = function(_, minStr, maxStr)
```
krryan: Newbie question: can we just use

```
	["(%d+) to (%d+) (%a) thorns damage"] = function(_, minStr, maxStr, typeStr)
		return {
			flag("GrantsThorns"),
			mod(typeStr .. "ThornsMin", "BASE", tonumber(minStr)),
			mod(typeStr .. "ThornsMax", "BASE", tonumber(maxStr)),
		}
	end,
```

here? I suppose there’s a risk of a non-(phys/cold/lightning/fire/chaos) entry here, but that seems small.

LocalIdentity (approved):
Want to change a bunch of the logic for this at some point to make it cleaner but will live with this for now
I added back the logic for eabling thorns only when you have a mod that grants it

PR #1782, opened 2026-03-24
Add support for Into the Breach purple flames

greatnameincoming:

Adds support for 

- Remnants have #% increased effect
- Effect and duration of chayula flames on you is doubled (legacy mod). The duration part of this line is ignored.

This has been applied to work on:

- Into the breach flames (from chayula monk)
- Life remnants (from bloodmage)
- Mana remnants

Only the purple breach flames are actually used in calculation, the others only show up in the Calcs tab.

- Verified remnant effect correctly scales the values of the above skills in the Calcs tab
- Verified the remnant effect modifier and breakdown is correct
- Verified the chaos damage applied by purple chayula flames is correct

https://pobb.in/edDxyBDYCfyV

### Before screenshot:

image

### After screenshot:

image

image

image


LocalIdentity (approved):
Fixed some logic so the buff doesn't duplicate with multiple gems and the chaos flames get disabled if you take either of the nodes that only spawns blue / red flames

PR #1783, opened 2026-03-25
Fix variant text ordering for The Adorned and Yoke of Suffering

### LINKED ISSUE #1774: The Adorned variant text is broken
The second line is duplicated on Current variant, and missing on 0.4.0. It shows the second line as variant 2 for both, changing it to 1 does not save. 
Applies to Yoke of Suffering too.
Image
Image
Image
Moving these lines so that it's in order fixes the issue. So an exporter change will probs fix this issue.
Image

poikende:
Multi-line mod continuation lines were interleaved between variants, causing duplicate/missing lines in the variant display. Group each variant's continuation lines together with their first line.

Fixes #1774

Fixes # .

The Adorned and Yoke of Suffering have multi-line mods where the continuation lines were interleaved between variants instead of grouped with their respective variant. This caused duplicate lines to appear for one variant and missing lines for the other in the item display.

- Loaded the items in PoB and switched between variants to confirm correct text displays for each

### Before screenshot:
image
image

### After screenshot:
image
image

### DIFF
--- a/src/Data/Uniques/amulet.lua
+++ b/src/Data/Uniques/amulet.lua
@@ -271,8 +271,8 @@ Implicits: 1
 {tags:fire,cold,lightning}+(10-15)% to all Elemental Resistances
 (15-30)% increased Elemental Damage
 {variant:1}Enemies take (5-10)% increased Damage for each Elemental Ailment type among
-{variant:2}Enemies take (15-20)% increased Damage for each Elemental Ailment type among
 {variant:1}your Ailments on them
+{variant:2}Enemies take (15-20)% increased Damage for each Elemental Ailment type among
 {variant:2}your Ailments on them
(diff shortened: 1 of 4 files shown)


Blitz54 (dismissed):
These changes will be overwritten any time we run the exporter. The proper way to fix this is in the uModsToText script. When I looked at it last, the issue is it sorts mod lines by the stat order, but for multiline mods on variants, we need to keep them bundled together.

Wires77:
Blitz is right, but thanks for calling this out.  I updated the script in your PR to properly do this for all uniques, even ones we don't parse properly.

Wires77 (approved):

poikende:
Sorry, that was my first open source pull request.
Makes sense that this issue affected more than 2 items and could be a sign of something broader. Nice of you to look into it, thanks.

PR #1796, opened 2026-04-15
Add a CI linter test for static code checks

Musholic:

This PR adds a new CI "linter" test to perform static code checking (to detect bugs) ~~as well as code style checks (to detect missing tabs, spaces, newlines, ...).~~ (I will do a second PR to deal with the code formatting)

Note that my initial commit uses the container "ghcr.io/musholic/lastepochplanner-tests:latest," which corresponds to the Dockerfile update I included. But I will revert that once it's proven (with a green CI at least) that the new Dockerfile update is good enough that we can release a new Docker build.

I've fixed most detected issues (at least the kind that I did not change the severity to "Information", these should be tackled in other PRs if wanted).

For the remaining issues I did not dare to fix myself, I added a `---@diagnostic disable` comment to make the CI test work until we fix them. It can be worth fixing them now, as they probably reflect real bugs.

Note that I'm using addons for luaLS (https://luals.github.io/wiki/addons/) for busted and luassert supports. They are supported in most IDEs (I suppose), like VScode for example, but it may be needed to disable then re-enable them in its addons manager (the addon manager only exists in VScode afaik, but once configured, it's available for other IDEs). Let me know if you have difficulties integrating with the IDE.

- [ ] Revert the container used for tests


vaisest:
Related pob1 PR: https://github.com/PathOfBuildingCommunity/PathOfBuilding/pull/9861

I'd say something of this sort would be borderline required before linting makes sense. The current code structure is very unfriendly to static analysis

PR #1807, opened 2026-04-28
Support opening builds from builds folder via command line argu…

C1rdec:
# Add Command-Line Support for Opening Local Builds

## Overview

This change adds support for opening existing local `.xml` builds via command-line arguments.

Previously, Path of Building only supported opening builds through the `pob2://` protocol, which is limited to importing builds from external sources. There was no way for external tools to open a build that already exists in the local builds folder.

## Usage

```bash
PathOfBuilding.exe "<path>/Path of Building (PoE2)/Builds/MyBuild.xml"
```

## Benefits

- Enables integration with external tools such as **PoeLurker**
- Enables **desktop shortcuts** to open specific builds directly

## Compatibility

- Existing `pob2://` protocol behavior is unchanged
- Launching without arguments behaves as before
- Invalid or missing files are handled gracefully (no crash)

## Testing

- ✅ Launch with valid `.xml` filename → build opens correctly  
- ✅ Launch with no arguments → default behavior unchanged  
- ✅ Launch with `pob2://` URI → protocol handler still works  
- ✅ Launch with invalid `.xml` filename → no crash, fails gracefully

### DIFF
--- a/src/Modules/Main.lua
+++ b/src/Modules/Main.lua
@@ -152,17 +152,17 @@ function main:Init()
 
 	-- Open a build by name from the builds folder (passed via command line)
 	if pendingBuildName and self.buildPath then
-		-- Ensure it ends with .xml and resolve relative to buildPath
-		local fileName = pendingBuildName:match("([^/]+)$")
-		if fileName then
-			local buildFile = self.buildPath .. fileName
-			if not buildFile:lower():match("%.xml$") then
-				buildFile = buildFile .. ".xml"
+		-- Reject path traversal attempts
+		if not pendingBuildName:match("%.%.") then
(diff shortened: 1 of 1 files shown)


Wires77:
Do people really set up shortcuts for a single build on their computer?  How is PoeLurker planning on implementing this?

C1rdec:
> Do people really set up shortcuts for a single build on their computer? How is PoeLurker planning on implementing this?

I’d probably use shortcuts myself, but for PoeLurker I’m currently listing PoB builds and linking a build to a character in an overlay, and the only way to open the Skill Tree is via the official website URL embedded in the XML. It would be really convenient if users could open the build directly in PoB.

PR #1809, opened 2026-04-30
Fix many skills not enabling the Wither config option

krryan:
The “# of Wither Stacks” config was not enabled by Chaos Bolt, His Foul Emergence, Withering Presence, Withering Touch, Wither Away, or The Right Hand of Darkness.

- Added each skill listed above one by one (without the others), to ensure the “# of Wither Stacks” config appeared as expected in each case.

https://pobb.in/RRWhH42PJV3X
(Not that there’s much point, just add any of the skills or take any of the nodes listed here.)

### Before screenshot:
image

### After screenshot:
image


krryan:
I attempted to add unit tests for this, but couldn’t figure out how to test that the config was shown. I was using

```
	it("Test Wither stack config with Chaos Bolt", function()
		build.skillsTab:PasteSocketGroup("Chaos Bolt 1/0  1\n")

		assert.is_true(build.configTab.varControls.multiplierWitheredStackCount.shown())
	end)
```

but I think I need to run something in between these two to get the `.shown()` to update. I tried `configTab:BuildModList()` and `configTab:UpdateControls()`, but those did not work. I realize there aren’t a ton of unit tests here, and this is a pretty simple thing that maybe doesn’t need them, but thought the commentary might be useful for maintainers as something to think about if you’d like to see more unit tests.

LocalIdentity (approved):

PR #1819, opened 2026-05-08
Fix increased Enemy damage taken mods doubling their affect in Ailment calcs

Andrew-Klotz:
Fixes bug with no outstanding issue.

Increases/more damage taken on enemies is double counted for physical/chaos ailments (bleed and poison). E.g., Nearby enemies take 10% increased chaos damage results in a 20% increase in the calculations.

Confirmed physical/chaos work properly for bleed/poison respectively, while elemental still works for ignite.

https://pobb.in/UCOPSE85Cyoz

### Before screenshot:

image

### After screenshot:

image

### DIFF
--- a/spec/System/TestAilments_spec.lua
+++ b/spec/System/TestAilments_spec.lua
@@ -14,4 +14,18 @@ describe("TestAilments", function()
 	--TODO: Shock not supported currently
 	--it("bleed is buffed by bleed chance", function()
 	--end)
+
+	it("does not double count chaos damage taken for chaos poison", function()
+		build.skillsTab:PasteSocketGroup("Chaos Bolt 1/0  1\nPoison I 1/0  1\n")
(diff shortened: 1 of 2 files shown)


Andrew-Klotz:
Looks good to me. I kept the ailmentTypeMod path as future proofing for other generic damage modifiers that could be added (e.g. non-chaos), but currently the game only has elemental so your cleaner fix works.

As for testing, didn't add any as I was unsure on testing philosophy (test count looks fairly light given the project size, and contributing.md just mentions adding testing for new features) and didn't know how much of a performance bottleneck there might be on testing. Depending on philosophy, it might be worth considering adding testing for increased phys for bleed and increased fire/elemental for ignite. Current dev has the same double-count for phys bleed as for chaos poison.

LocalIdentity (approved):

PR #1830, opened 2026-05-11
Add build comparison tab

vaisest:
Changes from https://github.com/PathOfBuildingCommunity/PathOfBuilding/pull/9543

Ports changes, but also improves the trader functionality a lot. Previously (and this is true for PoB1 too) any mod with separate global and local trade filters would fail to match correctly, which meant searching most armour or weapons was not working well. Inverted mods, which say increased on the item, but are reduced with the value negated on the trade site, were also not supported.

These are both now supported. I have tested quite a few items, but given how there are a gazillion stats, it's possible some of them don't function. There are for example some things like the 4 desecration jewel which lack the correct base tags. Megalomaniac is also not supported, because the trade hash format is weird.

Tested combinations of the following builds:

- Perfect strike (stages): https://pobb.in/5uT8UUi3wbrn
- Minion: https://pobb.in/4E2iq980M6Pi
- Ritualist bear (3 rings, weapon set): https://pobb.in/WdQrse5PFoMt
- Flicker (dps N/A): https://pobb.in/t3JF-mMMglzN
- Unique enjoyer (many uniques for trader tool): https://pobb.in/hblx1tFGBS1S

The stats have not been verified super accurately, but they seem correct, and nothing has crashed.

Seems most of it is directly compatible with pob2, but various things are hidden behind different variables. These fixes mostly focus on fixing those mistakes. Skill sets were also added to the calcs tab. 

Trade hash generation was also reworked. This probably means the trader tool will not generate QueryMods.lua correctly. I don't think anything else uses the tradehash field, though. The rework should correct A LOT of trade hashes. Many stats were being incorrectly concatenated, and most mods with multiple stats were wrong as concatenation is only correct for # to # mods which have a separate minimum and maximum stat


vaisest:
image

Increased passive skill effect now applies to tooltips (and this also fixes a crash), but it doesn't seem to be working for the power report lines. Not sure why.

LocalIdentity (approved):

PR #1853, opened 2026-05-13
Fix matchURL for BuildSites

Peechey:
When we ported over the URL spoofing PR, the matchURLs were set to "https:/" instead of "https://" This is not an issue in PoB1

### Before screenshot:
image

### After screenshot:
image


Wires77 (approved):
No idea how that happened, thanks

PR #1878, opened 2026-05-18
Default From Nothing imports to small radius

### LINKED ISSUE #1457: Onframe error
### Check version

### Check for duplicates

Windows

Hope this error to be solved.

When I import the "From Nothing" jewel from trade site, I encounter following "Onframe error"

1. import tree from character
2. copy "From Nothing Jewel" item from trade site
3. create custom item from copied jewel
4. import to the tree

### PoB for PoE2 build code

```shell

```

### Screenshots

Image

unrealdreamz:
## Summary
Fixes #1457

- defaults trade-site `From Nothing` jewel imports to the unique's known `Small` radius when the pasted text omits the `Radius: Small` property
- hardens `NodeInKeystoneRadius` so a missing or unavailable radius map returns false instead of indexing nil
- adds a regression fixture using the trade-site copied text from the issue

## Root Cause
Trade-site item text for `From Nothing` can omit the `Radius: Small` line that is present in the in-game copy. The parser only set `jewelRadiusIndex` from an explicit radius property, so the imported jewel kept its From Nothing keystone data but had no radius. Passive tree code later used that missing radius while checking keystone-radius allocation support, which could crash on nil indexing.

## Fix
After parsing jewel mods, `From Nothing` imports with no parsed radius now receive the generated unique's known `Small` radius. The keystone-radius helper also validates the radius before indexing tree radius maps, so malformed future jewel data fails closed rather than crashing.

## Validation
- `busted --lua=luajit ../spec/System/TestItemParse_spec.lua`
  - PASS: `25 successes / 0 failures / 0 errors / 0 pending : 0.281 seconds`
- `git diff --check`
  - PASS

## Risk/Rollback
Risk is limited to `From Nothing` jewel imports missing radius metadata and to safer nil handling in one passive tree radius helper. Rollback is this commit only.

### DIFF
--- a/spec/System/TestItemParse_spec.lua
+++ b/spec/System/TestItemParse_spec.lua
@@ -58,6 +58,26 @@ describe("TestItemParse", function()
 	--it("Sockets", function()
 	--end)
 
+
+	it("defaults trade-site From Nothing jewels to small radius", function()
+		local item = new("Item", [[
+			Rarity: Unique
+			From Nothing
+			Diamond
+			Item Level: 84
+			--------
+			Passives in Radius of Walker of the Wilds can be Allocated
+			without being connected to your tree
+			--------
+			Corrupted
+		]])
+
+		assert.are.equals("Small", item.jewelRadiusLabel)
+		assert.are.equals(1, item.jewelRadiusIndex)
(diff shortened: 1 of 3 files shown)


LocalIdentity:
No longer an issue as the website now shows the radius value

PR #1883, opened 2026-05-18
Fix item-granted skills counting toward socket group limit

### LINKED ISSUE #985: too many gem groups allocated, warning only
### Check version

### Check for duplicates

### Check for support

Windows

In game in "Skills" I am using 11 skills (Mace Strike, Raise Shield and 9 non-granted/non-item skills)

In PoB I see warning, see screenshot:

Image

_No response_

### PoB for PoE2 build code

```shell
https://maxroll.gg/poe2/pob/uk2i907u
```

### Screenshots

_No response_

unrealdreamz:
﻿Fixes #985

## Summary
- Give item-granted skill groups a stable generated source (`Item:<slot>`), matching existing tree-granted skills (`Tree:<node>`).
- Adopt older generated item skill groups that were saved with a nil source instead of creating a second group beside them.
- Treat generated item/tree skill groups as removable when their source no longer exists, while leaving ordinary user-created groups alone.
- Add a regression for a shield-granted Raise Shield group saved without a source.

## Root Cause
Item-granted skills were copied into `env.grantedSkillsItems` with `sourceItem` and `slotName`, but no stable `source`. The generated socket group was therefore created with `source = nil`. Later cleanup only removed groups with a truthy `socketGroup.source`, so stale generated item-skill groups could survive and duplicate when the granted-skill pass rebuilt groups. This is consistent with #985 reports involving extra generated skill groups and item-granted skills such as Raise Shield.

## Fix
The generated-skill pass now derives a stable source for item-granted skills, migrates old nil-source item-generated groups when they match the same slot/skill/level, and removes generated groups identified by `source`, `fromItem`, or `fromNode` when no current granted skill marks them.

## Validation
- Targeted regression added in `spec/System/TestSkills_spec.lua`.

## Risk/Rollback
Low-to-moderate lifecycle risk: the removal predicate is intentionally limited to generated skill groups (`source`, `fromItem`, or `fromNode`) so ordinary user-created groups without those markers are preserved. If an older save has an unusual generated group shape, rollback is a single CalcSetup change plus the regression removal.

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -246,44 +246,30 @@ describe("TestSkills", function()
 		assert.True(build.calcsTab.calcsOutput.Cooldown == 10)
 	end)
 
-	it("does not duplicate existing item-granted skill groups without a source", function()
-		build.itemsTab:CreateDisplayItemFromRaw([[
-			Rarity: UNIQUE
-			Chernobog's Pillar
-			Blacksteel Tower Shield
-			Implicits: 1
-			Grants Skill: Raise Shield
-			100% increased Armour
-			+30% to Fire Resistance
-			+23% to Chaos Resistance
-			+150 to Stun Threshold
-			Gain 1% of damage as Fire damage per 1% Chance to Block
-		]])
-		build.itemsTab:AddDisplayItem()
-		runCallback("OnFrame")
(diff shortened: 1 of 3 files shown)


LocalIdentity (approved):
I vastly simplified the fix

PR #1886, opened 2026-05-18
Fix exporter loading of old GGPK data

### LINKED ISSUE #1556: Fix issues loading old GGPK data in the exporter
### Check version

### Check for duplicates

Windows

When swapping the current DAT source in the exporter, we should be able to load old versions of GGPK data

When we changed the path from "Data" to "Data/Balance", we aren't able to load any older GGPKs because they don't have that path available.  In addition, I believe [this section](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/ee4128437629f0addacd6a00edefe7508a6e7ed9/src/Export/Main.lua#L380-L384) ends up overwriting the spec you're trying to load with the currently loaded one instead

Have a body of data saved from before 0.4.0, including spec.lua, data, metadata folders, etc.
Create a new Source in the exporter with that path, and switch to it to see no DAT files loaded.

### PoB for PoE2 build code

```shell

```

### Screenshots

Image

unrealdreamz:
﻿Fixes #1556

## Summary
- Let cached DAT loading find both current `Data/Balance/...` files and older pre-0.4.0 `Data/...` files.
- Save the currently selected DAT spec before switching sources, rather than writing the current spec into the newly selected source before loading it.
- Reuse one `SaveDatSpec` helper for source switching and shutdown.

## Root Cause
Two exporter paths were brittle for older GGPK data:
1. `AddDat64Files` only looked for cached files under the newer `Data/Balance` layout, so older extracted data under `Data` was reported missing.
2. `LoadDatSource` assigned `self.datSource = value` before saving, then wrote the previously loaded `self.datSpecs` to the newly selected source's spec path. That could overwrite the spec the user was trying to load.

## Fix
DAT lookup now tries the exact requested path first, then the legacy `Data/` path for `Data/Balance/` entries. Source switching now saves the previous source's spec before assigning and loading the new source.

## Validation
- No exporter UI/GGPK extraction run locally in this pass because this machine was explicitly kept free of local app/test windows.

## Risk/Rollback
Low exporter-only risk. Current cached data remains preferred because the new path is tried first; the legacy path is only a fallback. Rollback is isolated to exporter source/spec loading and cached DAT lookup.

### DIFF
--- a/src/Export/Classes/GGPKData.lua
+++ b/src/Export/Classes/GGPKData.lua
@@ -29,6 +29,16 @@ local function scanDir(directory, extension)
 	return t
 end
 
+local function getDatFilePathCandidates(fname)
+	if fname:match("^Data/Balance/") then
+		return {
+			fname,
+			fname:gsub("^Data/Balance/", "Data/"),
+		}
+	end
+	return { fname }
+end
+
 -- Path can be in any format recognized by the extractor at oozPath, ie,
 -- a .ggpk file or a Steam Path of Exile directory
 local GGPKClass = newClass("GGPKData", function(self, path, datPath, reExport)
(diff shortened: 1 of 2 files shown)


LocalIdentity:
Not needed anymore

PR #1887, opened 2026-05-18
Handle trade responses without rate limit headers

### LINKED ISSUE #506: Error: In download callback: Classes/TradeQueryRateLimiter.lua:76 table index is nil
### Check version

### Check for duplicates

Windows

### Screenshots of crash/graphical issue, version and options menu

Image

### What are your system specifications and configuration?

Win11,
Run as an administrator.

Select Items -> Trade for these items
**Note that the SSL connection error message is displayed**
When I used version 0.2.0 yesterday, this feature worked fine.

### Character build code

```shell

```

unrealdreamz:
﻿Fixes #506

## Summary
- Treat trade responses without `X-Rate-Limit-Policy` as non-policy responses instead of indexing nil.
- Preserve existing rate-limit state when Cloudflare/SSL/HTML error responses lack trade rate-limit headers.
- Avoid clearing `POESESSID` unless a real rate-limit rules header is present and lacks `Account`.
- Add regression coverage for missing policy headers and missing-policy updates.

## Root Cause
The trade callback assumed every response had Path of Exile trade rate-limit headers. SSL/Cloudflare/error responses can contain only an HTTP status and generic headers, so `ParsePolicy` attempted to index `policies[nil]`, and the request callback chained `:match(...):match(...)` through a nil match.

## Fix
`ParsePolicy` now returns an empty policy table when the policy header is absent, parses `Retry-After` numerically, and ignores malformed buckets. The request callback now normalises the response header to an empty string, updates rate limits only when parseable headers exist, and only performs the invalid-session check when the `X-Rate-Limit-Rules` header exists.

## Validation
- Targeted regressions added in `spec/System/TestTradeQueryRateLimiter_spec.lua`.

## Risk/Rollback
Low trade-request risk: valid policy headers still parse through the existing path. The changed behavior is limited to missing/malformed policy responses, where crashing or deleting session state was worse than preserving current limits and surfacing the original request error.

### DIFF
--- a/spec/System/TestTradeQueryRateLimiter_spec.lua
+++ b/spec/System/TestTradeQueryRateLimiter_spec.lua
@@ -24,6 +24,12 @@ describe("TradeQueryRateLimiter", function()
 			assert.are.equal(policy.ip.state[10].request, 7)
 			assert.are.equal(policy.account.limits[5].request, 2)
 		end)
+
+		it("ignores responses without rate limit policy headers", function()
+			local limiter = new("TradeQueryRateLimiter")
+			local policies = limiter:ParsePolicy("HTTP/1.1 503 Service Unavailable\nContent-Type: text/html")
+			assert.are.same({}, policies)
+		end)
 	end)
 
 	describe("UpdateFromHeader", function()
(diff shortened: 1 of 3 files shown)


LocalIdentity:
No longer relevant as we now use OAuth once the next update comes out

PR #1888, opened 2026-05-18
Fix applyRange plus sign at zero

### LINKED ISSUE #1412: Error when changing slider value of Edonia's Ego Maximum Power Charges Mod
For some mods that go between a negative value and a positive, we get an error when going from positive to 0. Adonia's Ego with `{range:0.5}+(-1-1) to Maximum Power Charges` is the easiest way to test. Changing the numbers to -5-5 still bugs, but higher values like -50-50 there is no error when landing on 0.
It's something in itemLib.applyRange. Replacing that func with the one in PoB1 removes the error. But of course we need the new code for corrupted mod scalability.

Image

unrealdreamz:
﻿Fixes #1412

## Summary
- Stop preserving a leading `+` when ranged item text resolves exactly to zero.
- Add an `applyRange` regression for a signed range centered on zero.

## Root Cause
`itemLib.applyRange` preserved the original plus sign when the interpolated range value was `>= 0`. That produced `+0 ...` for ranges that land exactly on zero, even though the rest of item text formatting uses plain `0`.

## Fix
Preserve `+` only for strictly positive interpolated values. Negative values already drop the plus sign, and now zero does too.

## Validation
- `git diff --check` - pass, exit code 0.
- Targeted regression added in `spec/System/TestItemTools_spec.lua`.

## Risk/Rollback
Very low risk: the change is limited to signed range formatting when an originally plus-prefixed range resolves to exactly zero.

### DIFF
--- a/spec/System/TestItemTools_spec.lua
+++ b/spec/System/TestItemTools_spec.lua
@@ -36,6 +36,9 @@ local applyRangeTests = {
 	[{ "+(-25-50)% to Fire Resistance", 1.0, 1.5 }] = "+75% to Fire Resistance",
 	[{ "+(-25-50)% to Fire Resistance", 0.0, 1.0 }] = "-25% to Fire Resistance",
 	[{ "+(-25-50)% to Fire Resistance", 0.0, 1.5 }] = "-37% to Fire Resistance",
+
+	-- Range with plus sign that resolves exactly to zero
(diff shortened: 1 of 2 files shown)


LocalIdentity (changes requested):
Does not work, currently just hides the slider now

Issue #983, opened 2025-04-12
Leopold's Applause reduced enemy lightning resistance minimum from penetration is not being applied


zrisher:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

In-game, when I have Leopold's Applause equipped, my lightning penetration should be able to reduce enemy resistances to -50% instead of 0%.

### What is the value from the calculation in Path of Building?

In PoB, when I have Leopold's Applause equipped and view my MH Eff. DPS Mod on the Calcs screen, I see Effective Resistance to be 0%.

### How to reproduce the issue

Equip Leopold's Applause
Give yourself > 50% lightning pen
View your MH Eff DPS Mod

### PoB for PoE2 build code

```shell
https://poe2.ninja/pob/2865
```

### Screenshots

![Image](https://github.com/user-attachments/assets/7dc1af2b-a1c4-4a6f-8dfd-30848089dc19)

![Image](https://github.com/user-attachments/assets/479fdf12-24fc-4ace-b4c5-6b3fff7af977)

Blitz54:
![Image](https://github.com/user-attachments/assets/98fe23b7-ef0e-4607-85e6-b5435446afba)
Red text is a feature request. Expected to not work yet as we haven't implemented it yet.

zrisher:
Ah I see. I didn't know what "red text" this line was referring to:

> I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

Perhaps adding "(pending)" or "(not yet implemented)" to the end of the red text would be more instructive to people that do look on the item and see this issue, though.

Do I need to open a feature request instead?

SingleMalt23:
[...]

PR #1895, opened 2026-05-18
Fix penetration minimum DPS breakdown

unrealdreamz:
## Summary
Refs #983

Make the Calcs breakdown use the same penetration floor as the hit calculation when showing the Effective DPS modifier.

## Root Cause
The main hit calculation already supports `ElementalPenetrationMinimum`, including Leopold's Applause allowing elemental penetration down to `-50%`. The breakdown generator still displayed the resistance multiplier with a hard `0%` floor:

```lua
math.max(resist - pen, 0)
```

That made the Calcs explanation show a `1.00` resistance component even when the actual calculation used negative effective resistance.

## Fix
- Default `minPen` to `0` in `breakdown.effMult` for existing callers.
- Compute the displayed effective resistance with the same rule used by the offence calculation:
  - if resistance is already below the penetration floor, leave it alone;
  - otherwise clamp penetration at `minPen`.
- Use that value in the Effective DPS modifier breakdown.
- Add direct breakdown regressions for Leopold-style `-50%` penetration floor and normal negative resistance without penetration.

## Validation
- `git diff --check` passed.
- `git diff --cached --check` passed.

Not run locally:
- `docker-compose up` / full Busted suite. The repository test runner is Docker-based; I avoided local container/window activity and left the focused regression for CI.

## Risk / Rollback
Low risk. This changes only explanatory breakdown output, not the damage calculation itself. Rollback is the single commit on this branch.

### DIFF
--- a/spec/System/TestCalcBreakdown_spec.lua
+++ b/spec/System/TestCalcBreakdown_spec.lua
@@ -1,0 +1,20 @@
+describe("TestCalcBreakdown", function()
+	local calcBreakdown = LoadModule("Modules/CalcBreakdown", {}, {}, {})
+
+	it("shows penetration minimum in effective DPS breakdown", function()
+		local out = calcBreakdown.effMult("Lightning", 50, 100, 0, 1.5, 1, nil, true, nil, -50)
+		local text = table.concat(out, "\n")
+
+		assert.True(text:match("= %-50%%") ~= nil)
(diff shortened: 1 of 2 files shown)


LocalIdentity:
Already fixed by https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/pull/1136

PR #1900, opened 2026-05-18
Fix active gem DPS sorting

### LINKED ISSUE #1633: Active Gems Contributing to DPS but not sorted for DPS
### Check version

### Check for duplicates

Windows

If an active gem contributes to DPS it should have the green plus next to it like some other gems do.

Currently on a crossbow build Oil Grenade has a green plus because it contributes in my case 700k to Combined DPS which I sort for. But explosive grenade doesn't have the plus even if it contributes double the amount.

(obviously I spiked the custom modifier to check if there was perhaps a min value under which it doesnt get a green plus)

1. Add a crossbow to mainhand
2. Go to gems tab
3. Sort by Combined DPS
4. In search click A for active gems
5. Notice 1 skill with green plus out of the 12 that actually do.

### PoB for PoE2 build code

```shell
https://pobb.in/7AYn7ekqNYSs
```

### Screenshots

Image

unrealdreamz:
﻿## Summary

Fixes #1633.

Includes active replacement gems in the existing DPS sort estimate so active gems that change the selected skill's DPS are ranked by their actual impact instead of being left at the base build DPS.

## Root Cause

`GemSelectControl:UpdateSortCache` only ran the DPS comparison for support gems that could support the displayed active skill, plus non-support gems with global effects. Ordinary active gems in the active-gem slot were still valid replacement candidates, but they were never sent through `CalcOutputWithThisGem`, so they kept the base DPS value and missed the impact marker.

## Fix

- Track which gem candidates actually receive a DPS estimate.
- Estimate DPS for non-support gems when the selector is editing the active gem slot.
- Show the existing `+` impact marker only for active replacement gems whose calculated DPS differs from the base DPS.
- Add a regression covering an Oil Grenade active slot where Explosive Grenade must receive a calculated DPS entry.

## Validation

- `git diff --check` — passed, CRLF conversion warnings only.
- `git diff --cached --check` — passed, CRLF conversion warnings only.
- `git show --check --stat --oneline HEAD` — passed.

I did not run Docker/Busted locally on this machine; the local test path for this repo is container-based and I avoided starting containers or external windows during this contribution pass.

## Risk / Rollback

Risk is limited to the gem selector DPS-sort cache. The change reuses the existing calculator path already used for support/global-effect gem comparisons, and only broadens it for active-gem replacements in the active slot. Rollback is the single commit if it causes selector-performance or sorting regressions.

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -245,4 +245,25 @@ describe("TestSkills", function()
 
 		assert.True(build.calcsTab.calcsOutput.Cooldown == 10)
 	end)
+
+	it("Test active gem DPS sorting estimates replacement skills", function()
+		build.itemsTab:CreateDisplayItemFromRaw([[
+			New Item
+			Bombard Crossbow
+			Quality: 0
+		]])
+		build.itemsTab:AddDisplayItem()
+		runCallback("OnFrame")
+
+		build.skillsTab:PasteSocketGroup("Oil Grenade 20/0  1")
+		runCallback("OnFrame")
+
+		build.skillsTab.sortGemsByDPS = true
+		build.skillsTab.sortGemsByDPSField = "CombinedDPS"
+		local activeGemSelect = build.skillsTab.gemSlots[1].nameSpec
+		activeGemSelect:UpdateSortCache()
+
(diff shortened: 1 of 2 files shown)


LocalIdentity:
Fixing this requires iterating across stat sets and have configs set correctly and even then it only applies for the first skill in a build so is largely useless

PR #1904, opened 2026-05-18
Fix weapon set allocation issues + add colour to nodes

### LINKED ISSUE #1532: Weapon Set Passives – Invalid Allocation
### Check version

### Check for duplicates

### Check for support

- [x] I've checked that the behaviour is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

Windows

### What is the behaviour in-game?

You cannot allocate a normal passive point at the end of a weapon set passive branch.

### What is the behaviour in Path of Building?

PoB allows you to allocate normal passives at the end of a weapon set passive branch.

The first normal passive you add at the end of the branch will also appear as the same color as the weapon set even though it's being added as a normal passive.

1. Set entry mode as weapon set I or II
2. Allocate passives on a connected part of the tree
3. Set entry mode to normal passives
4. Continue adding points after the weapon set points

### PoB for PoE2 build code

```shell
https://pobb.in/d-d5qyyOf2Uw
```

### Screenshots

Image

unrealdreamz:
﻿## Summary

Prevent normal passive allocation from continuing through weapon-set-only branches.

Fixes #1532

## Root Cause

Passive path construction treats every allocated node as a path root without remembering which root produced a candidate path. That means a node at the end of a weapon-set branch can become the closest root for the next unallocated passive. When allocation mode is switched back to normal, `AllocNode` only checks that a path exists, so it can allocate a normal passive from a weapon-set-only branch.

## Fix

- Track the allocated root that produced each passive path via `pathRoot`.
- Clear `pathRoot` whenever paths are rebuilt, alongside `path`/`pathDist`.
- In normal allocation mode, reject paths rooted in an allocated weapon-set node or explicitly passing through one.
- Leave weapon-set mode allocation unchanged.
- Added a focused passive-spec regression that:
  - allocates one normal node,
  - allocates the next node in weapon set 1,
  - proves normal mode cannot continue beyond that branch,
  - proves weapon set 1 still can.

## Validation

Local validation run:

```text
git diff --check
PASS (CRLF normalization warnings only)

git diff --cached --check
PASS (CRLF normalization warnings only)

git show --check --stat --oneline HEAD
PASS (CRLF normalization warnings only)
```

I did not run Docker/Busted locally because the available full test runner is container-based and I was avoiding external test containers/windows during this session.

## Risk / Rollback

Risk is localized to passive allocation/path bookkeeping. The guard applies only while normal allocation mode is selected; weapon-set mode continues to use the existing path behavior. Rollback is the single commit if an unexpected valid normal-path edge case appears.

### DIFF
--- a/spec/System/TestPassiveSpec_spec.lua
+++ b/spec/System/TestPassiveSpec_spec.lua
@@ -1,0 +1,71 @@
+describe("TestPassiveSpec", function()
+	before_each(function()
+		newBuild()
+	end)
+
+	local function pathContains(path, needle)
+		if not path then
+			return false
+		end
+		for _, node in ipairs(path) do
+			if node == needle then
+				return true
+			end
+		end
+		return false
+	end
+
+	local function isAllocatableNormalNode(node)
+		return node
+			and not node.alloc
+			and node.type == "Normal"
+			and not node.ascendancyName
+			and not node.isMultipleChoice
+			and not node.isMultipleChoiceOption
+			and #node.intuitiveLeapLikesAffecting == 0
+	end
+
+	local function findNormalPathNodeAfter(spec, throughNode)
+		for _, node in ipairs(throughNode.linked) do
+			if isAllocatableNormalNode(node) and (node.pathRoot == throughNode or pathContains(node.path, throughNode)) then
+				return node
+			end
+		end
+		for _, node in pairs(spec.nodes) do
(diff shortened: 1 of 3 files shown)


LocalIdentity (approved):
I fixed the logic for this as there were many mistakes and edge cases that were not covered

Issue #714, opened 2025-02-05
Infernal Cry Empowered Attack Support


JotCGit:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the behaviour is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the behaviour in-game?

In-game, Infernal Cry empowers my next attack based off of the power of enemies.

### What is the behaviour in Path of Building?

In PoB2, under "Configuration", there is a checkbox for "Is the skill Empowered?" which doesn't show up even if I have Infernal Cry enabled and checked for "Full DPS". Even if I show "Ineligible Configurations" and check the empowered box, nothing happens to my DPS. Furthermore, even if I adjust the "# of Warcries Used Recently" specification and the "Warcry Power" specification, nothing happens to my DPS either.

### How to reproduce the issue

1. Add skill Infernal Cry.
2. Add skill "Perfect Strike"
3. Have both "Included in Full DPS"
4. Check to see if the DPS is affected with infernal cry enabled or disabled.

### PoB for PoE2 build code

```shell
I'm not a coder I have no idea what code to produce here.
```

bicben:
I can confirm it does not affect the DPS no matter what. 100% bonus fire damage is not there

bwoolmer:
This is still an issue.

You can manually type the damage gain in config however it would be ideal if the "Skill is empowered" checkbox worked.

Weirdly enough. When removing the custom config. The combust moves into the Full DPS??


![Image](https://github.com/user-attachments/assets/f26958c8-0f1a-41d4-983d-2b497a79fdf1)

![Image](https://github.com/user-attachments/assets/1b2ca81d-5d12-4279-b951-c6904fb49acd)

![Image](https://github.com/user-attachments/assets/9f82961e-c5e8-4893-a781-9558d2205625)

![Image](https://github.com/user-attachments/assets/d28c139d-7a36-46f2-8f67-cb86e0bb690b)

[...]

PR #1911, opened 2026-05-18
Apply Infernal Cry extra fire to empowered attacks

unrealdreamz:
﻿## Summary

Refs #714

Adds the missing stat mapping for Infernal Cry's empowered attack bonus so its `Gain x% of Damage as Extra Fire Damage` stat is emitted as a Warcry buff for empowered melee attacks.

## Root Cause

Infernal Cry's exported skill data already included `infernal_cry_exerted_attack_all_damage_%_to_gain_as_fire_%`, and the stat description rendered it for users, but the skill had no `statMap` entry. The stat therefore never became a live modifier, so enabling Infernal Cry could show no extra fire damage effect on empowered melee attacks.

This PR intentionally does not claim to finish the broader automatic Full DPS / Warcry Power behavior discussed in #714; it fixes the missing damage modifier path only.

## Fix

- Add a generated-data and export-source `statMap` entry for Infernal Cry.
- Emit `DamageGainAsFire` as a melee-only Warcry global effect.
- Gate the modifier behind the existing `Empowered` condition so non-empowered attacks are not affected.
- Add a system regression that verifies Infernal Cry now produces a Warcry buff modifier with the melee flag and Empowered condition tag.

## Validation

- `gh pr list --repo PathOfBuildingCommunity/PathOfBuilding-PoE2 --state open -S '714 Infernal Cry empowered fire' --json number,title,headRefName,author,url --jq '.'` -> `[]`
- `git diff --check` -> passed
- `git diff --cached --check` -> passed
- `git show --check --stat --oneline HEAD` -> passed

I did not run Docker/Busted locally in this pass because this machine is currently in active foreground use; the regression is included for CI to run in the repository's normal test container.

## Risk / Rollback

Low risk. The change is data-scoped to Infernal Cry and applies only when the active skill is an empowered melee attack. Rollback is reverting this commit, which removes the stat mapping and the regression test.

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -229,6 +229,40 @@ describe("TestSkills", function()
 		assert.True(baseCorruptingCryDps == build.calcsTab.mainOutput.CorruptingBloodDPS)
 	end)
 
+	it("Infernal Cry grants extra fire damage to empowered melee attacks", function()
+		build.skillsTab:PasteSocketGroup("Infernal Cry 20/0  1")
+		runCallback("OnFrame")
+
+		local infernalCry
+		for _, activeSkill in ipairs(build.calcsTab.mainEnv.player.activeSkillList) do
+			if activeSkill.activeEffect.grantedEffect.name == "Infernal Cry" then
+				infernalCry = activeSkill
+				break
+			end
+		end
+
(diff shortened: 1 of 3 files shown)


LocalIdentity:
Does not correctly implement the buff as an uptime based on the number of empowered attacks. Need to use the same exert system as PoB 1

Issue #783, opened 2025-02-14
Bow Attacks not calculating "Bow Attacks fire an Additional Arrow" or "Bow Attacks fire 2 Additional Arrows"


John-The-Tripper:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

The calculation is also missing in game from my research. For example, I have a Dual-string bow with suffix to fire another arrow for a total of 4 arrows. The Lightning Arrow skill information shows that it's only firing 2. I believe this issue is both game client side and in PoB.

### What is the value from the calculation in Path of Building?

The damage per second value is not modified by adjusting a bow from only firing one additional arrow to two. In my test case I fire 4 arrows and an upgrade would see me firing 5 arrows for an increase of 20% DPS (or 25%, I suck at math). But in testing and manually modifying an imported Bow from PoE2 I've found no change to the damage profile.

### How to reproduce the issue

1. Ranger with Deadeye Ascendency and "Endless Munitions"
2. Dual-string Bow with Suffix "Bow Attacks Fire an Additional Arrow"
3. Copy bow to PoB2
4. Edit bow to have Suffix "Bow Attacks Fire 2 Additional Arrows" and remove "Bow Attacks Fire an Additional Arrow"

### PoB for PoE2 build code

```shell
[...]

PR #1914, opened 2026-05-18
Apply bow additional arrows to bow attacks

unrealdreamz:
﻿## Summary

Refs #783.

Makes `Bow Attacks fire an additional Arrow` / `Bow Attacks fire 2 additional Arrows` apply to bow attacks' projectile count.

This intentionally does not multiply normal displayed hit DPS by arrow count. It only fixes the objective parser/matching issue that prevented the projectile-count modifier from matching current PoE2 bow skill flags.

## Root Cause

The parser mapped `Bow Attacks fire ... additional Arrows` to `KeywordFlag.Arrow`. Current bow attacks such as Lightning Arrow receive `ModFlag.Bow` / bow weapon matching, but do not appear to set `skillFlags.arrow`, so these mods were filtered out before `ProjectileCount` was summed.

## Fix

- Changed the bow-specific additional-arrow parser entries to use `ModFlag.Bow`.
- Kept generic arrow/projectile entries unchanged.
- Added a regression test proving the custom line increases Lightning Arrow projectile count with a bow equipped and does not affect a non-bow projectile spell.

## Validation

Local, non-GUI validation only:

```text
gh pr list --repo PathOfBuildingCommunity/PathOfBuilding-PoE2 --state open -S "783 bow attacks additional arrows projectile count" -> []
git diff --check -> passed
git diff --cached --check -> passed
git show --check --stat --oneline HEAD -> passed
where.exe lua -> not found
where.exe luajit -> not found
where.exe busted -> not found
```

I did not run Busted locally because this Windows checkout does not have Lua/LuaJIT/Busted on PATH. I also avoided Docker/GUI validation so no visible command windows or heavy processes were launched.

## Risk / Rollback

Low risk. The change is limited to parser entries that explicitly say "Bow Attacks" and narrows matching to bow attacks via the existing `ModFlag.Bow` path. Rollback is this single commit.


LocalIdentity:
user issue so this PR does not apply

Issue #387, opened 2025-01-19
Rage count from global configuration not accounted for in HOTG damage calculation


sanson87:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

I hope this is relevant since I am not an expert in the game mechanics. Currently when I add a rage count buff in global configuration like this :

![Image](https://github.com/user-attachments/assets/3b7e8fbf-2593-46a1-b0e0-130e7b0a4212)

I can see it accounted for in Mace strike damage calculation:

![Image](https://github.com/user-attachments/assets/8ee8785c-8fa9-4020-a490-d91f5fd2e17e)

But not for HOTG : 

![Image](https://github.com/user-attachments/assets/0f724222-418c-439e-87b0-5ab3c5966e37)

### What is the value from the calculation in Path of Building?

The rage buff should be accounted for just like for Mace strike in this example (I haven't tested other melee attack skills)

### How to reproduce the issue

1. Equip hammer of the gods along with whatever mace
2. Add a rage count to the global configuration
3. Go to calculation and check for the MH Total More value while "In combat" mode is selected.

### Character build code

```shell
[...]

PR #1917, opened 2026-05-18
Recognize Ngamahu's Chosen Rage source

unrealdreamz:
﻿## Summary
- Recognise `Grants up to your maximum Rage on use` as a Rage-enabling modifier.
- Add a regression proving configured Rage still requires a recognised Rage source.
- Add a regression proving the Ngamahu's Chosen wording enables configured Rage and increases attack damage through the existing Rage calculation path.

Refs #387

## Root Cause
The Rage calculation intentionally requires a recognised Rage source before configured Rage stacks are applied. That invariant is important because otherwise a stale or manually-entered Rage value could silently grant impossible attack damage.

The later #387 reproduction uses Ngamahu's Chosen. Its unique modifier, `Grants up to your maximum Rage on use`, was not recognised by the modifier parser, so the build did not get `Condition:CanGainRage`; configured Rage then stayed inactive even though the item is a valid Rage source.

## Fix
- Add a parser entry for `Grants up to your maximum Rage on use` that emits `Condition:CanGainRage`.
- Leave the core Rage gate unchanged: configured Rage still applies only when the build has a recognised Rage source or Rage regeneration.
- Cover both sides in tests: no source means no Rage output/damage increase; the Ngamahu wording means configured Rage is applied.

## Validation
- `gh pr list --repo PathOfBuildingCommunity/PathOfBuilding-PoE2 --state open -S '387 rage configured CanGainRage Ngamahu' --json number,title,headRefName,author,url --jq '.'` -> `[]` before PR creation
- `git diff --check` -> passed
- `git diff --cached --check` -> passed before commit/amend
- `git show --check --stat --oneline HEAD` -> passed
- `where/Get-Command lua` -> not found on PATH
- `where/Get-Command luajit` -> not found on PATH
- `where/Get-Command busted` -> not found on PATH
- `docker --version` / `docker-compose --version` -> not found on PATH

Targeted Busted coverage was added in `spec/System/TestSkills_spec.lua`, but I could not execute it locally because this checkout does not have Lua/LuaJIT/Busted or Docker available on PATH.

## Risk / Rollback
Risk is narrow: one exact Rage-source parser line plus two tests. The core Rage calculation and config semantics are unchanged.

Rollback is the single commit on this branch.

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -193,60 +193,6 @@ describe("TestSkills", function()
 		assert.True(baseLeapSlamHit < build.calcsTab.mainOutput.AverageDamage)
 	end)
 
-	it("Test configured Rage still requires a recognised Rage source", function()
-		build.itemsTab:CreateDisplayItemFromRaw([[
-			New Item
-			Fanatic Greathammer
-			Quality: 0
-		]])
-		build.itemsTab:AddDisplayItem()
-		runCallback("OnFrame")
-
-		build.skillsTab:PasteSocketGroup("Leap Slam 20/0  1")
-		runCallback("OnFrame")
-
-		local baseLeapSlamHit = build.calcsTab.mainOutput.AverageDamage
-
-		build.configTab.input.multiplierRage = 30
-		build.configTab:BuildModList()
-		runCallback("OnFrame")
-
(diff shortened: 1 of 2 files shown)


LocalIdentity:
This does not work correctly and charm specific mods need to be handled in a different way for this rage mod to work

Issue #261, opened 2025-01-17
Shockwave totem is severely undercalculated


chx:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

https://imgur.com/a/STumeI6

### What is the value from the calculation in Path of Building?

![Image](https://github.com/user-attachments/assets/deabd4f0-f222-40ff-b0c5-f82cf8af227e)

### Character build code

```shell
[...]

PR #1918, opened 2026-05-18
Use totem actor level for totem life

unrealdreamz:
﻿## Summary

Refs #261

Use the totem skill's exported actor level when calculating totem level/life, with the previous minion-gem-level table kept as the fallback for totem sources that do not export an actor level.

## Root Cause

Totem life is calculated from `monsterAllyLifeTable[skillData.totemLevel]`, but `skillData.totemLevel` was always derived from `data.minionLevelTable[gemLevel]`.

For Shockwave Totem, the exported skill stat set gives level 20 an actor level of about 97.7. The minion level table maps gem level 20 to level 40, so the totem life calculation used level 40 monster life instead of the skill's own exported actor level.

## Fix

- Add a small `getTotemLevel` helper beside the existing totem base-skill resolution.
- Prefer `actorLevel` from the totem granted effect/stat set when present.
- Round and clamp the result to the available monster ally life table.
- Preserve the old `data.minionLevelTable` behavior as the fallback for totem sources without exported actor levels.
- Add a regression spec for level 20 Shockwave Totem, asserting its totem level resolves to 98 and its displayed totem life uses `monsterAllyLifeTable[98]`.

## Validation

- `gh pr list --repo PathOfBuildingCommunity/PathOfBuilding-PoE2 --state open -S 'Shockwave Totem actor level totem life' --json number,title,headRefName,author,url --jq '.'` -> `[]`
- `gh pr list --repo PathOfBuildingCommunity/PathOfBuilding-PoE2 --state open -S '261 Shockwave Totem' --json number,title,headRefName,author,url --jq '.'` -> `[]`
- `git diff --check` -> passed (Git emitted only local CRLF conversion warnings)
- `git diff --cached --check` -> passed (Git emitted only local CRLF conversion warnings)
- `git show --check --oneline --no-renames HEAD` -> passed
- `Get-Command lua`, `luajit`, `busted`, `docker`, `docker-compose` -> not found on this Windows PATH, so I could not execute the Busted suite locally here.

## Risk / Rollback

Risk is limited to totem level/life calculations. Totems that do not export actor levels keep the previous minion-level fallback. Rollback is the single commit on this branch.

### DIFF
--- a/spec/System/TestTotems_spec.lua
+++ b/spec/System/TestTotems_spec.lua
@@ -1,0 +1,25 @@
+describe("TestTotems", function()
+	before_each(function()
+		newBuild()
+	end)
+
+	teardown(function()
+		-- newBuild() takes care of resetting everything in setup()
+	end)
+
+	it("uses exported actor level for Shockwave Totem life", function()
+		build.skillsTab:PasteSocketGroup("Shockwave Totem 20/0  1")
+		runCallback("OnFrame")
+
+		local checkedTotems = 0
+		for _, activeSkill in ipairs(build.calcsTab.calcsEnv.player.activeSkillList) do
+			local totemBase = activeSkill.skillData.totemBase
+			if totemBase and totemBase.grantedEffect and totemBase.grantedEffect.name == "Shockwave Totem" then
+				checkedTotems = checkedTotems + 1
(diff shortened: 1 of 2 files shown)


LocalIdentity:
Not a problem anymore and the totem uses the players weapon

Issue #258, opened 2025-01-17
Arcane Surge adds too much cast speed, mana regen, possible solution to calculation, or just bug in-game


Iamproplayer7:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

Without Arcane Surge in-game: 
![Image](https://github.com/user-attachments/assets/1e8009e8-47ed-4a07-8350-004cc1599604)

With Arcane Surge in-game:
![Image](https://github.com/user-attachments/assets/72163886-3f30-4af5-b8dd-f474ae05ab06)

### What is the value from the calculation in Path of Building?

Without Arcane Surge in-pob: 
![Image](https://github.com/user-attachments/assets/3fb4284e-9fa7-4f4a-9836-4e6c3e28e855)

With Arcane Surge in-pob:
![Image](https://github.com/user-attachments/assets/2bd4f6ed-7434-4af8-9202-1f52c732e461)
![Image](https://github.com/user-attachments/assets/6d937d15-e57a-4fba-8022-31e91acb5b1d)

### How to reproduce the issue

Remove all those nodes in game: 
![Image](https://github.com/user-attachments/assets/a6a819e7-4404-4393-8778-5aca70d745e8)

Remove all those nodes in pob:
![Image](https://github.com/user-attachments/assets/810c03dd-8801-413e-91e6-bf4d2f7a3f94)

### Character build code

```shell
[...]

PR #1919, opened 2026-05-18
Scale Force of Will by missing mana

unrealdreamz:
﻿Refs #258

## Summary
- Scale Force of Will's Arcane Surge effect by missing mana instead of treating the full line as a flat 20% increased effect.
- Add a missing-mana multiplier derived from the existing Current Mana % config value.
- Add a regression spec for the exact passive text.

## Root Cause
The modifier parser recognized `20% increased Effect of Arcane Surge on you` and left `per ten percent missing Mana` as unparsed trailing text in `ModCache.lua`. That made Force of Will contribute a flat 20% Arcane Surge effect regardless of current mana.

## Fix
- Added a special parser rule for `increased effect of arcane surge on you per ten percent missing mana`.
- Added `Multiplier:MissingManaPercentage` beside the existing clamped `Multiplier:CurrentManaPercentage` config value.
- Tagged the parsed mod with `div = 10`, so it uses full ten-percent chunks through the existing multiplier semantics.
- Updated the generated cache entry for this exact modifier and added a fixture-style regression in `TestItemMods_spec.lua`.

## Validation
- `gh auth status` -> authenticated as `unrealdreamz`.
- Open PR search for this bug family -> no duplicate open PR found.
- `git diff --check` -> passed; Git only reported existing Windows LF-to-CRLF checkout warnings.
- `git diff --cached --check` -> passed; Git only reported existing Windows LF-to-CRLF checkout warnings.
- `git show --check --stat --oneline --no-renames HEAD` -> passed.
- Targeted Busted spec was added but could not be executed locally because this Windows environment has no `lua`, `luajit`, `busted`, `docker`, or `docker-compose` on PATH.

## Risk / Rollback
Risk is limited to modifiers that explicitly use the new missing-mana multiplier and to the existing Current Mana % config path. Rollback is the single commit if CI finds a parser/cache mismatch.

### DIFF
--- a/spec/System/TestItemMods_spec.lua
+++ b/spec/System/TestItemMods_spec.lua
@@ -221,6 +221,25 @@ describe("TetsItemMods", function()
 		runCallback("OnFrame")
 	end)
 
+	it("Arcane Surge effect per ten percent missing mana", function()
+		build.configTab.input.customMods = [[
+			20% increased Effect of Arcane Surge on you per ten percent missing Mana
+		]]
+		build.configTab.input.multiplierCurrentManaPercentage = 100
+		build.configTab:BuildModList()
+		runCallback("OnFrame")
+
+		assert.equals(0, build.calcsTab.mainEnv.modDB:GetMultiplier("MissingManaPercentage"))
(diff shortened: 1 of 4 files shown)


LocalIdentity:
Bug with the game, not PoB

Issue #280, opened 2025-01-18
Witchhunter Sorcery Ward doesn't apply to ward


Jung-o:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding/issues?q=is%3Aissue)

### Check for support

- [x] I've checked that the calculation is supposed to be supported. If it isn't please open a feature request instead (Red text is a feature request).

### What platform are you running Path of Building on?

Windows

### What is the value from the calculation in-game?

In game armour is 3300 and evasion is 8659.
Sorcery Ward is 5382

### What is the value from the calculation in Path of Building?

In PoB, armour is 3024 and evasion is 7456, no idea where the diff comes from but doesn't matter in this issue.
Sorcery Ward doesn't appear in the ward category in calculations and doesn't impact EHP

### Character build code

```shell
[...]

PR #1927, opened 2026-05-19
Support Sorcery Ward aegis calculation

unrealdreamz:
﻿## Summary

Refs #280

Adds calculation support for Sorcery Ward's barrier value from armour and evasion, including Sorcery Ward effect scaling and the all-damage barrier variant.

## Root Cause

The exported Sorcery Ward skill data already includes `aegis_unique_shield_max_value_from_%_armour_evasion = 30`, but `SkillStatMap` had no mapping for that stat. As a result, the granted Sorcery Ward skill could exist without producing an aegis pool for defence/EHP calculations. The related Sorcery Ward effect and all-damage barrier passive lines also were not parsed.

## Fix

- Map Sorcery Ward's armour/evasion percentage stat into an aegis buff:
  - elemental aegis by default
  - shared aegis when the all-damage barrier condition is present
- Parse increased/reduced Sorcery Ward effect so ascendancy passives scale the barrier value.
- Parse `Sorcery Ward's Barrier can also take Physical and Chaos Damage from Hits` as the condition that switches the barrier from elemental-only aegis to shared aegis.
- Add focused defence specs for both elemental-only and all-damage Sorcery Ward behaviour.

## Validation

Commands run locally:

```text
PASS lua compile src/Data/SkillStatMap.lua
PASS lua compile src/Modules/ModParser.lua
PASS lua compile spec/System/TestDefence_spec.lua

git diff --check
# passed

git diff --cached --check
# passed

git show --check --stat --oneline --no-renames HEAD
# passed; commit 417803bcd Support Sorcery Ward aegis calculation
```

Tooling unavailable in this checkout/host, so the targeted Busted spec and Docker suite could not be executed locally:

```text
lua=NOT FOUND
luajit=NOT FOUND
busted=NOT FOUND
docker=NOT FOUND
docker-compose=NOT FOUND
```

## Risk / Rollback

Low-to-moderate risk. The new stat mapping is scoped to Sorcery Ward's exported aegis stat and the new parser entries are exact Sorcery Ward lines. If the aegis model needs adjustment, rollback is just the three touched files: `src/Data/SkillStatMap.lua`, `src/Modules/ModParser.lua`, and `spec/System/TestDefence_spec.lua`.

### DIFF
--- a/spec/System/TestDefence_spec.lua
+++ b/spec/System/TestDefence_spec.lua
@@ -30,6 +30,40 @@ describe("TestDefence", function()
 		return build.calcsTab.calcs.reducePoolsByDamage(nil, takenDamages, build.calcsTab.calcsEnv.player)
 	end
 
+	it("Sorcery Ward grants elemental aegis from armour and evasion", function()
+		build.configTab.input.enemyIsBoss = "None"
+		build.configTab.input.customMods = "\z
+		+10000 to Armour\n\z
+		+5000 to Evasion Rating\n\z
+		50% increased effect of Sorcery Ward\n\z
+		"
+		build.configTab:BuildModList()
+		build.skillsTab:PasteSocketGroup("Sorcery Ward 1/0  1")
+		runCallback("OnFrame")
+
+		local output = build.calcsTab.calcsOutput
+		local expectedAegis = math.ceil((output.Armour + output.Evasion) * 0.45)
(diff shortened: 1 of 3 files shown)


LocalIdentity:
Does not correctly use a different guard type and have correct breakdowns
Does not apply the sorcery ward effect mod to the buff

PR #1935, opened 2026-05-22
[0.5] Add legacy toggle for gems that are no longer obtainable

Peechey:
Gems cleanup based on https://www.poe2wiki.net/wiki/Category:Historical_game_content and the "Can no longer be obtained" updates in the 0.5 patch notes https://www.pathofexile.com/forum/view-thread/3932540

### After screenshot:

image

image


LocalIdentity (approved):

src/Export/Skills/sup_dex.txt
```diff
@@ -797,16 +753,6 @@ statMap = {
 #mods
 #skillEnd
 
-#skill SupportOverextendPlayer
```
LocalIdentity, line 802: Should be marked as legacy instead of removed completely

src/Export/Skills/sup_str.txt
```diff
@@ -1006,22 +976,6 @@ statMap = {
 #mods
 #skillEnd
 
-#skill SupportInfernalLegionPlayerThree
```
LocalIdentity, line 977: Should be marked as legacy

src/Export/Skills/sup_int.txt
```diff
@@ -1343,11 +1310,6 @@ statMap = {
 #mods
 #skillEnd
 
-#skill SupportShockConductionPlayerTwo
```
LocalIdentity, line 1363: Should be marked as legacy

LocalIdentity (changes requested):
These 3 in the 0.5 patch notes only say they cannot be obtained, not that they are disabled
Not sure about the other gems if they also should be getting the legacy tag

LocalIdentity (approved):

PR #1937, opened 2026-05-22
[0.5] Update Deflection formula

Peechey:
`The formula for chance to Deflect has been adjusted to provide better scaling with investment into Deflection Rating, with a cap of 95% chance to Deflect (similar to chance to Evade being capped at 95%).`

`The new chance to Deflect is = 150*(1 - A/(A + 0.12*D)), where A = attacker accuracy rating, and D = defender deflection rating. This makes for a linear path, that gets rewarded at higher investments. `

### DIFF
--- a/src/Modules/CalcDefence.lua
+++ b/src/Modules/CalcDefence.lua
@@ -49,8 +49,8 @@ function calcs.deflectChance(deflection, accuracy)
 	if deflection < 1 then
 		return 0
 	end
-	local rawChance = ( accuracy * 0.9 ) / ( accuracy + deflection * 0.2 ) * 100
-	return 100 - m_max(m_min(round(rawChance), 100), 0)
+	local rawChance = accuracy / ( accuracy + deflection * 0.12 ) * 150
+	return 100 - m_max(m_min(round(rawChance), data.misc.DeflectionChanceCap), 0)
 end
 -- Calculate damage reduction from armour, float
 function calcs.armourReductionF(armour, raw)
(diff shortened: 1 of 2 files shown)


src/Modules/CalcDefence.lua
```diff
@@ -49,8 +49,8 @@ function calcs.deflectChance(deflection, accuracy)
 	if deflection < 1 then
 		return 0
 	end
-	local rawChance = ( accuracy * 0.9 ) / ( accuracy + deflection * 0.2 ) * 100
-	return 100 - m_max(m_min(round(rawChance), 100), 0)
+	local rawChance = accuracy / ( accuracy + deflection * 0.12 ) * 150
+	return 100 - m_max(m_min(round(rawChance), data.misc.DeflectionChanceCap), 0)
```
Edvinas-Smita: That's my bad on poor variable naming - the first calculation is chance to not deflect and since that chance is then subtracted from 100 that would end up with the updated formula being `100-150*A/(A+D*0.12)` instead of `150-150*A/(A + 0.12*D)`

```suggestion
	local chanceToNotDeflect = accuracy / ( accuracy + deflection * 0.12 ) * 150 - 50
	return 100 - m_max(m_min(round(chanceToNotDeflect), data.misc.DeflectionChanceCap), 0)
```

The suggested change should fix that and also probably good to correct the variable name.

Edvinas-Smita (changes requested):

Peechey:
Maintainers, give Edvinas credit in the patch notes for this one or else 😠

LocalIdentity (approved):

PR #1947, opened 2026-05-23
Support flicker strike -related mod lines, and various support gems

vaisest:

Adds support for:
- 20% more critical hit chance quality line on flicker strike
- flicker strike burst damage, dps display, and the 285% more AS line
- heightened charges
- enables armour break config for uul-netol support
- fixes blazing critica, which had support but didn't work
- hand of chayula mark effect
- elemental conflux by defaulting to an average of each element, while having a config option for a single element
- one with the storm

- tests for conflux and flicker charge behaviour
- manual check of each gem
- hollow palm checked

These are just the red lines I had on my 0.4 build: https://pobb.in/Lp8iPSAfTF3v

### After screenshot:
image
image
image


vaisest:
Charged staff + heightened charges is not properly working in this PR, but seems to be fixed by #1928. This adds the scalar in a similar way, and it should work when merged with that PR

LocalIdentity (approved):

PR #1980, opened 2026-05-26
Add AI assistant drawer for build analysis

wangdong3213:
## Summary

Adds an optional OpenAI-compatible AI assistant drawer to the build editor. The drawer appears on the right side of an open build, can be collapsed, and provides build analysis or follow-up answers without modifying the build.

## What Changed

- Added a right-side `AIChatPanel` for build analysis chat.
- Added local-only AI settings stored in `AIAssistantSettings.json` under the user data path.
- Added a minimal non-streaming OpenAI-compatible chat completions client.
- Added an English build snapshot generator covering character info, active sets, skill groups, key stats, items, and warnings.
- Integrated the drawer into `Build.lua` with small lifecycle insertions while preserving existing tabs, including Compare.

## Notes

- This is English-only and does not depend on PoeCharm2 or localization files.
- The assistant gives advice only; it does not write to build XML or change passives, gear, skills, or config.
- API keys are stored locally only and are not included in exported/shared builds.
- Streaming/realtime reasoning display is intentionally out of scope.

## Validation

- `rg` check confirmed no PoeCharm2 or Chinese UI/prompt strings remain in the new AI code.
- `git diff --check`
- Windows headless startup via `runtime/Path{space}of{space}Building-PoE2.exe src/HeadlessWrapper.lua`
- Targeted module load check for `AI/ProviderPresets`, `AI/Settings`, `AI/AIClient`, `AI/BuildSnapshot`, and `Classes/AIChatPanel`

Local Docker/LuaJIT were not installed, so the full Busted workflow was left to GitHub CI.

### DIFF
--- a/src/AI/AIClient.lua
+++ b/src/AI/AIClient.lua
@@ -1,0 +1,65 @@
+-- Path of Building
+--
+-- Module: AI Client
+-- Minimal non-streaming OpenAI-compatible chat completions client.
+
+local json = require("dkjson")
+
+local client = { }
+
+local function endpointFromBase(apiBase)
+	apiBase = (apiBase or ""):gsub("%s+$", "")
+	if apiBase:match("/chat/completions$") then
+		return apiBase
+	end
+	apiBase = apiBase:gsub("/+$", "")
+	return apiBase .. "/chat/completions"
+end
+
+local function extractContent(responseText)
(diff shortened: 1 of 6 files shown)


LocalIdentity:
We are not interested in adding AI assistants into Path of Building as we do not believe to be healthy for users and for the game

PR #1983, opened 2026-05-27
Pin calc stats to sidebar

### LINKED ISSUE #1731: User defined UI element positions & parent -> child nesting in custom sub-categories for the calculations screen
### Check for duplicates

Windows

This is a convenience & accessibility suggestion to clean up the sometimes cluttered and hard to read at a glance calculations menu.

Essentially, an additional function that allows the user to add a categorical, parent menu to add specific, titled windows within the calculations screen for which the user can configure where stats will be positioned categorised / outlined seperately from others _(i.e. Defence, Attack, DOT, EHP, etc)_.

**Drag/drop & "+ to add category menu" button** would be nice, but even just a config file that makes doing this possible would be awesome.

_No response_

### Additional context

Situationally, I find myself chasing a single stat window in calculations amongst a tightly packed bunch of others and I get lost quick if I'm looking between the game and POB. It would be nice to have a setup for example, poison / bleed calcs + whatever else contributes in their own specific spaces to see more clearly at a glance.

KenManDev:
Fixes #1731 

The Calcs tab can be difficult to read at a glance when trying to monitor a small number of specific values. This adds a lightweight way to keep selected calculation values visible without requiring a custom layout system.

- Added the ability to pin/unpin Calcs tab values to the left sidebar by right-clicking them.
- Left-click behavior is unchanged and still pins/preserves the breakdown popup within calcs tab.
- Added a "Clear Pinned Stats" button next to the Calcs tab search field.
- Added helper text explaining that right-clicking a stat pins it to the sidebar.

- Verified that right-clicking a Calcs tab value adds it to the pinned section in the sidebar.
- Verified that right-clicking the same value again removes it from the pinned section.
- Verified that the pinned sidebar values use the existing calc formatting and update correctly.
- Verified that left-clicking still preserves the existing breakdown popup behavior.
- Verified that the "Clear Pinned Stats" button removes all pinned values.

### Before screenshot:
image

### After screenshot:
image

### DIFF
--- a/src/Classes/CalcSectionControl.lua
+++ b/src/Classes/CalcSectionControl.lua
@@ -74,6 +74,7 @@ function CalcSectionClass:IsMouseOver()
 							return mOver, colData
 						end
 					elseif cursorX >= colData.x and cursorY >= colData.y and cursorX < colData.x + colData.width and cursorY < colData.y + colData.height then
+							colData.pinnedLabel = data.label
 						return mOver, colData
 					end
 				end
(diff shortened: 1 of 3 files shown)


LocalIdentity:
No longer required as you can now pin whole breakdown sections and the sidebar also shows breakdowns on hover

PR #1991, opened 2026-05-27
Add Martial Artist Rune slots

vaisest:

This adds runic meridians support by:
- importing runes from json
- adding dropdown controls (which primarily save the rune name, no dummy items created so not an ItemSlotControl)
- saving runes to weapon sets
- saving and loading them in builds

This also adds a warning for when you exceed the augment limit for any given augment.

- testing with monk builds from poe.ninja
- playing with my own character which is not a monk (nothing has broken)
- some tests were added, although build save/load or import are not tested as these didn't seem to be tested anywhere else either.

https://poe.ninja/poe2/builds/runesofaldur/character/heygyus-0416/ResurrectGodAura?i=0&search=class%3DMartial%2BArtist

### After screenshot:
image
image


src/Modules/ModParser.lua
```diff
@@ -6103,7 +6103,10 @@ local specialModList = {
 	} end,
 	["you can socket an additional copy of each lineage support gem, in different skills"] = { mod("MaxLineageCount", "BASE", 1) },
 	["you can socket (%d+) additional copies of each lineage support gem, in different skills"] = function(num) return { mod("MaxLineageCount", "BASE", num) } end,
-	["can be modified while corrupted"] = {}
+	["can be modified while corrupted"] = {},
+	["can tattoo runes onto your body, gaining"] = { flag("SocketRunesOnCharacter") },
```
vaisest, line 6233: Not really sure if this is the correct way to do this. It would probably be possible to parse the socket counts, but this seems much simpler

jerrybeanman:
Hello. I'm pretty new to the repo. I noticed that the runic meridian support is missing in the current release and tracked down this PR. Wondering if there's an ETA for when we can see this in since it's pretty crucial for a lot of the MA calculations, especially with a lot of builds socketing Aldur's legacy there? Thank you as always you guys are awesome!  :)

@LocalIdentity @vaisest is there any way to move this forward and get it in for the next release?

vaisest:
> Hello. I'm pretty new to the repo. I noticed that the runic meridian support is missing in the current release and tracked down this PR. Wondering if there's an ETA for when we can expect to see this in since it's pretty crucial for a lot of the MA calculations, especially with a lot of builds socketing Aldur's legacy there? Thank you!
> 
> @@ is there any way to move this forward and get it in for the next release?

I don't know when this will be merged as it's not up to me. Either way it's certainly not crucial. You can use the custom mods box on the config tab to apply these effects and it should work just the same for now.

jerrybeanman:
> > Hello. I'm pretty new to the repo. I noticed that the runic meridian support is missing in the current release and tracked down this PR. Wondering if there's an ETA for when we can expect to see this in since it's pretty crucial for a lot of the MA calculations, especially with a lot of builds socketing Aldur's legacy there? Thank you!
> > @@ is there any way to move this forward and get it in for the next release?
> 
> I don't know when this will be merged as it's not up to me. Either way it's certainly not crucial. You can use the custom mods box on the config tab to apply these effects and it should work just the same for now.

Thank you for taking the time to respond. Yes I have been doing that for the moment. This feature would save a lot of time as it is a bit tedious to manually add in the runes whenever I need to reimport my character. Also would be a big plus to have the numbers reflected on ninja as well. Great work as always

LocalIdentity (approved):

PR #2008, opened 2026-05-29
Fix compare colors

Divinicus1st:
- Fix added nodes in compare mode correctly appearing green
- Added/Removed node pathing in compare mode changed to teal/pink to distinguish from weapon set 1-2 point pathing (in green/red)

### Before screenshot:

image

### After screenshot:

image


Divinicus1st:
edit: fixed the "after" screenshot

PR #2027, opened 2026-05-31
Update quest rewards for 0.5

rasmuskl:

0.5 API now serves `quest_stats` as aggregated instead of exact lines per quest. So the +30/30/40 lines for spirit becomes a single +100 line instead. This PR adds a simple greedy algorithm to configure quests in import by decomposing the rolled up values.

Adds a simple disambiguation table for spirit specifically, since the 30+30+40 spirit values can result in wrong values with the greedy algorithm. An alternative solution would be to try all combinations, but I decided to keep it simple since it was only a single case to avoid the additional complexity.

Also added the new Tribal Medicine quest: https://poe2db.tw/us/Tribal_Medicine

Added new tests and tested in UI.


Wires77 (approved):
Annoying change, but thanks for the quick fix and tests

PR #2046, opened 2026-06-01
Fix Comet not showing DPS values

LocalIdentity:
The skill has the only show average damage flag on it so I gave it a blank stat map to stop it from hiding the DPS calcs

https://maxroll.gg/poe2/pob/k06x0d0y

### Before screenshot:
image

### After screenshot:
image


Wires77:
Why only this one and not just change the behavior for all of the average skills?

PR #2049, opened 2026-06-01
Change tests to run in parallel + cache cleanup

LocalIdentity:
The tests before ran in one run without a global cache cleanup
This change cuts the peak memory usage from 1.23GB to 940MB which hopefully is enough for the tests to stop failing


Wires77 (approved):
I might iterate on this to use busted's tags, but this looks fine otherwise

PR #2058, opened 2026-06-02
fix: correct Runic Ward import, double-count, and EHP operator precedence

jay9297:
## Summary

- **`ImportTab:ImportItem` missing `return item`** — The function parsed the GGG API item correctly but never returned it, so all API-import callers received `nil`. Items imported via the character API were silently dropped.
- **Ward double-counting on authoritative items** — When `armourData.Ward` is pre-set from a paste or API property line (`Runic Ward: 104`), that value is the game's final post-quality result. The old code still applied `wardInc` rune mods on top of the already-final value, and left unconsumed flat BASE mods to leak into `CalcDefence`. Fixed with a `wardIsAuthoritative` flag: both BASE and INC local mods are consumed (removed from modList) but discarded; the property-line value is used verbatim.
- **Operator precedence zeroing EHP pool** — `output[...TotalHitPool] + output.Ward or 0` parsed as `(... + output.Ward) or 0`, silently zeroing the entire hit pool when Ward was nil. Fixed with explicit parentheses: `(output.Ward or 0)`.

## Test plan

- [ ] All 392 existing tests pass: `docker compose run --rm tests`
- [ ] Paste a Runeforged item with `Runic Ward: X` — displayed Ward equals `X` (no quality re-scaling)
- [ ] Import a character via GGG API wearing a Runeforged body armour — item appears in the Items tab with correct Ward value
- [ ] Character with Ward equipped shows correct EHP (Life + Ward, not zeroed)
- [ ] Item with rune mods in `runeModLines` — `#runeModLines` reflects rune count, mods not double-applied


jay9297:
Opened against wrong repo — re-submitting against fork.

PR #2097, opened 2026-06-05
Add Facebreaker unique gloves support

Anakior:
 Adds Facebreaker and support since i'm playing it for its three modifiers that currently show up as  "Not supported in PoB yet":
- `Has 8 to 12 Physical damage, +3 to +4 per Boss's Face Broken` (also the Fire variant)  
- `Can Attack as though using a One Handed Mace while both of your hand slots are empty`
- `Unarmed Attacks that would use an Equipped One Hand Mace's damage use this Item's damage`

The "attack as a One Hand Mace while unarmed" bit works like Hollow Palm: when both hands are empty I set `weaponData1.asThoughUsing["One Hand Mace"]` in CalcSetup so mace skills become usable. The hits stay unarmed-flagged (it only satisfies the weapon requirement, it doesn't grant mace scaling). The item's base damage is added as flat unarmed damage.

The `per Boss's Face Broken` part is a `Multiplier:BossFaceBroken` driven by a new config option ("# of Boss's Faces Broken"), since that count isn't exposed by the character API and has to be entered manually like other "per X" stacks.

Also added the unique to Data + Export, regenerated ModCache, and added tests in `spec/System/TestFacebreaker_spec.lua`.

### Screenshot

image

image

### DIFF
--- a/spec/System/TestFacebreaker_spec.lua
+++ b/spec/System/TestFacebreaker_spec.lua
@@ -89,6 +89,11 @@ describe("TestFacebreaker", function()
 		assert.are.equals(0, modDB:Sum("MORE", { flags = ModFlag.Sword + ModFlag.Hit }, "Damage"))
 	end)
 
+	it("auto-imports the Boss's Faces Broken count from character quest stats", function()
+		build.importTab:ImportQuestRewardConfig({ "57 [BrokenFace|Broken Boss Faces]" })
+		assert.are.equals(57, build.configTab.input.configBossFaceBroken)
+	end)
+
 	it("supports the Fire damage variant", function()
 		build.itemsTab:CreateDisplayItemFromRaw([[
 			New Item
--- a/spec/System/TestFacebreaker_spec.lua
+++ b/spec/System/TestFacebreaker_spec.lua
@@ -1,0 +1,106 @@
+-- Tests for Facebreaker-style gloves: empty-handed gloves that grant their own base
+-- weapon damage and let you attack as though using a One Hand Mace.
+describe("TestFacebreaker", function()
+	before_each(function()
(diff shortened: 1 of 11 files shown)


vaisest:
> The per Boss's Face Broken part is a Multiplier:BossFaceBroken driven by a new config option ("# of Boss's Faces Broken"), since that count isn't exposed by the character API and has to be entered manually like other "per X" stacks.

This is incorrect. The API does have this info in the quest field 

Edit:

```json
      "quest_stats": [
        "\u002B20 to maximum Life",
        "5% increased maximum Life",
        "5% increased maximum Mana",
        "\u002B100 to [Spirit|Spirit]",
        "\u002B5% to all [ElementalDamage|Elemental] [Resistances]",
        "\u002B15% to [Resistances|Fire Resistance]",
        "\u002B15% to [Resistances|Cold Resistance]",
        "\u002B15% to [Resistances|Lightning Resistance]",
        "30% increased Life Recovery from [Flask|Flasks]",
        "30% increased Global [Armour], [Evasion] and [EnergyShield|Energy Shield]",
        "25% increased [StunThreshold|Stun Threshold]",
        "30% increased [Charm] Charges gained",
        "\u002B1 [Charm] Slot",
        "57 [BrokenFace|Broken Boss Faces]"
      ]
```

Anakior:
Oh ok. I didn't saw that field, thanks a lot ! 

I have updated the branch to get it from this and auto fill the config but didn't removed it from the config if people want to theorycraft on it.

LocalIdentity (approved):
I fixed some of the implementation so that it now works as a base damage source vs added damage
It now correctly works with hollow palm too

PR #2118, opened 2026-06-07
Buy similar: combine same stats from different mods

vaisest:

Fixes trade search for repeated mods with the same stats. For example breath of aldur weapons or items with resistances transformed to chaos res.

# to # mods now also use the midpoint for the search.

- Tested with various items from the trade site

https://pobb.in/xLtX_75reIyy
### Before screenshot:
image

The resulting trade search was also wrong as it only takes into account the first filter for each stat.

### After screenshot:
image
image
image


vaisest:
While taking screenshots I noticed the fields floor float values and didn't accept floats. This seems a bit misleading for e.g. flat crit so in the last commit I made mod fields float valued. Many mods use integer values, but even if the user inputs a float, it shouldn't affect anything since the trade site seems to floor them (though don't quote me on that)

LocalIdentity (approved):

PR #2121, opened 2026-06-07
Add trader support for Heart of the Well, Against the Darkness, Desecrated mods, Rune-only mods and Breach tree mods

vaisest:

This adds to the trader:
- Heart of the Well
- Against the Darkness
- Desecrated mods (many of them were just regular mods, but desecrate-only mods are now also present)
- Rune influenced mods (can roll X runes) (Meta mods like increased stat magnitudes will still do nothing (though they ARE taken into account as the trade site uses the old format which includes the effect in the resulting item text))
- Breach belt and ring mods

- Testing with my build
- Some specifics like a couple breach mods were also tested manually

### Before screenshot:
image

### After screenshot:

Panel widened slightly:

image

image


vaisest:
One issue I found while testing this PR was that the weights can be very confusing for minion builds. By default the idea is to use your stats, but if minions do all of your damage and you don't have full DPS set up, your dps weight will be zero.

Not sure what the best approach here is. I imagine if both you and your minions do damage, it's not a good idea to only use one of them

LocalIdentity (approved):

PR #2125, opened 2026-06-07
Add Bloodbarrier support

DS-Koala:
Fixes Bloodbarrier.

- Bloodbarrier was not supported as a calculated skill in Path of Building 2.
- Once created as an ExtraSkill, it still would not show up in Full DPS. Added a function `isSkillIncludedInFullDPS` to Calcs.lua that will work with extra/generated skills.

- Added Bloodbarrier to a build and confirmed it appears in the Skills tab.
- Confirmed Bloodbarrier calculates its Corrupted Blood damage correctly in the skill breakdown.
- Checked "Show in Full DPS" for Bloodbarrier and confirmed the damage is included in Full DPS.

https://pobb.in/b-PE53yEnGTe

### Before screenshot:
image

### After screenshot:
image


LocalIdentity (approved):
I fxed a bunch of issues with the PR
There was not export of the skill from other.txt
The gem.lua file is autogenerated from game data so you can't add to it
The FullDPS logic was not needed as PoB already handles corrupted blood DPS and only applies the strongest one
The SkillGems.txt file is auto-generated from game data, so it should not be edited
The formatting of the ModParser lines was not in keeping with the existing layout and in the wrong position

PR #2135, opened 2026-06-08
Add Predator's Mark support (Pounce's ally-based DamageTaken curse)

mataide:
Predator's Mark (`WolfPounceMarkPlayer`), the hidden mark triggered by Pounce, had no PoB support. The skill causes the marked enemy to take increased damage based on the number of allies near it, but the stat was never mapped, no config option existed, and the gem tooltip showed the stat line as unsupported.

This PR adds full support for Predator's Mark:
- **`statMap`** in `WolfPounceMarkPlayer`: maps `skill_wolf_mark_damage_taken_+%_per_nearby_enemy` to `DamageTaken INC (Multiplier:AlliesInSummonerRange, GlobalEffect:Curse)`, and marks the cap stat as supported so the tooltip renders in green
- **Config option** `# of Allies in Range:` (appears in Skill Options when Pounce is active; uses `ifEnemyMult = "AlliesInSummonerRange"`)
- **`CalcPerform` handler**: reads the config value, caps it at 15 (since cap/rate = 45/3 = 15), and writes to `enemyDB.multipliers["AlliesInSummonerRange"]`; also sets `output.PredatorsMarkAllies` for the CalcSections row
- **CalcSections row**: `Predator's Mark Allies` row in the offence section, visible when Pounce is active and allies > 0

- [x] Config option `# of Allies in Range:` appears in Skill Options when Pounce is in the skill list, hidden otherwise
- [x] Setting allies increases enemy `DamageTaken INC` in CalcBreakdown (3% per ally at base, scales with quality)
- [x] Cap enforced: setting 15 and 20 allies produces identical output
- [x] `Predator's Mark Allies` row is visible in the calcsTab offence section when Pounce is active and allies > 0
- [x] Gem tooltip stat line "Marked target takes X% increased damage for each of its enemies within 6 metres of it, up to Y%" renders in green (supported)
- [x] Predator's Mark config option does not affect builds without Pounce
- [x] Added automated tests to `TestSkills_spec.lua` covering the scaling and cap

Add a Pounce gem to any Druid build, set `# of Allies in Range: 5` in the Config tab → Skill Options, and observe the DamageTaken increase in the CalcBreakdown.

### DIFF
--- a/src/Data/Skills/act_str.lua
+++ b/src/Data/Skills/act_str.lua
@@ -15211,9 +15211,11 @@ skills["WolfPounceMarkPlayer"] = {
 			},
 			statMap = {
 				["skill_wolf_mark_damage_taken_+%_per_nearby_enemy"] = {
-					mod("DamageTaken", "INC", nil, 0, 0, { type = "Multiplier", var = "MinionPresenceCount" }, { type = "GlobalEffect", effectType = "Curse" }),
+					mod("DamageTaken", "INC", nil, 0, 0, { type = "Multiplier", var = "MinionPresenceCount", actor = "player", limitVar = "PredatorsMarkLimit", limitTotal = true }, { type = "GlobalEffect", effectType = "Curse" }),
 				},
-				["skill_wolf_mark_damage_taken_+%_cap"] = {},
+				["skill_wolf_mark_damage_taken_+%_cap"] = {
+					mod("Multiplier:PredatorsMarkLimit", "BASE", nil, 0, 0, { type = "GlobalEffect", effectType = "Curse" }),
+				},
 			},
 			constantStats = {
 				{ "base_skill_effect_duration", 8000 },
--- a/src/Data/Skills/act_str.lua
+++ b/src/Data/Skills/act_str.lua
@@ -15215,6 +15215,7 @@ skills["WolfPounceMarkPlayer"] = {
 						{ type = "Multiplier", var = "AlliesInSummonerRange" },
 						{ type = "GlobalEffect", effectType = "Curse" }),
 				},
+				["skill_wolf_mark_damage_taken_+%_cap"] = {},
 			},
 			constantStats = {
 				{ "base_skill_effect_duration", 8000 },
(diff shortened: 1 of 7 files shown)


src/Data/Skills/act_str.lua
```diff
@@ -15209,6 +15209,14 @@ skills["WolfPounceMarkPlayer"] = {
 			baseFlags = {
 				duration = true,
 			},
+			statMap = {
+				["skill_wolf_mark_damage_taken_+%_per_nearby_enemy"] = {
```
LocalIdentity, line 15213: Formatting of the mod is not consistent with other mods where we have it all on 1 line

src/Modules/CalcSections.lua
```diff
@@ -865,10 +865,12 @@ return {
 		{ breakdown = "ParryRangeNonProj" },
 		{ label = "Range modifiers", modName = "ParryRangeNonProj", cfg = "skill" },
 	}, },
-	{ label = "Parry Range Proj", haveOutput = "ParryRangeProj", { format = "{1:output:ParryRangeProj}m", 
-		{ breakdown = "ParryRangeProj" }, 
+	{ label = "Parry Range Proj", haveOutput = "ParryRangeProj", { format = "{1:output:ParryRangeProj}m",
+		{ breakdown = "ParryRangeProj" },
 		{ label = "Range modifiers", modName = "ParryRangeProj", cfg = "skill" },
 	}, },
+	-- Predator's Mark
```
LocalIdentity: This stat doesn't need to exist imo. Maybe one for MinionPresenceCount would be ok though

LocalIdentity (changes requested):
We already calculate MinionPresenceCount in CalcPerform so that value should be used for the mod

LocalIdentity:
Changes to the lua gem files also need to have the same change in the txt file that generates them

LocalIdentity (approved):
Please test your PRs and provide a build code + images. I had to fix the mod not working at all and also it was not using the damage cap mod to cap the stat
If you're not sure about how to do something, just ask in Discord and I'm happy to help

mataide:
@LocalIdentity Thanks for the help, after the change to MinionPresenceCount I still had to test.

PR #2140, opened 2026-06-09
Add missing genesis tree exclusive mods to Rings and Belts

ltogniolli:
…elt and ring bases

These mods only spawn on the genesis_tree_* tags, which GGG grants at runtime and omits from base data, so they matched no item. Adding the tags to belt/ring bases during export lets the existing spawn-weight logic route them to the right slot.

Couldn't add minion/caster mods to rings and belts

Check the existence of mods when crafting belts and rings. 

### Before screenshot:

image

### After screenshot:

image

### DIFF
--- a/src/Data/Bases/belt.lua
+++ b/src/Data/Bases/belt.lua
@@ -5,7 +5,7 @@ local itemBases = ...
 itemBases["Golden Obi"] = {
 	type = "Belt",
 	charmLimit = 0,
-	tags = { belt = true, default = true, demigods = true, not_for_sale = true, },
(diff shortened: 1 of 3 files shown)


LocalIdentity (changes requested):
These mods should not be included in the base mod pool and instead should be something needed to be added via `Add Modifier`

vaisest:
By the way, I think not including mods in the base pool causes issue with mod parsing.

For example the case where desecrated mods turn into invalid monk glove mods I mentioned earlier seemed to be caused by the parser matching against `data.itemMods.Item`, which only includes the base pool. I would imagine the same problem occurs here if they're not included in that table

I.e.

1. Import item with mod
2. Move any slider 
3. Mod gets replaced with another one because it's id was parsed wrong

LocalIdentity (approved):

PR #2146, opened 2026-06-10
Add support for Runic ward

Blitz54:

Adds Ward into PoB2. Lots of the old code still functioned thankfully.

- Ward regeneration shows on the left side bar, by default it's 5% per second.
- Ward is changed to last line of defence, in PoE2 it takes damage after life.
- Fixes parsing of Runic Ward on items.
- Fixes Armour, ES, and EV rune applying to Ward too.
- Runic Ward costs for skills
- Runic Ward cost efficiency
- Runic Ward cost warning when not enough to use skill
- Runic Ward on block
- Add support for Scouring Flame and other supports using same stat
- Add support for Runic Infusion
- Add support for Olroth's Hubris
- Also fixes armour values on tooltip not printing in blue at times that they should.
- Rune of Equinox - Mana regen also applies to ward

image
image


src/Modules/CalcDefence.lua
```diff
@@ -3897,6 +3910,7 @@ function calcs.buildDefenceEstimations(env, actor)
 					end
 					totalLifeDegen = totalLifeDegen + lifeDegen
 					totalManaDegen = totalManaDegen + manaDegen
+					totalWanaDegen = totalWanaDegen + wardDegen
```
Nightblade: typo?  should this be `totalWardDegen`?

LocalIdentity (approved):

PR #2147, opened 2026-06-10
Add support for tamed beast modifiers on Companion gems

ltogniolli:
Rare tamed beasts carry randomly rolled monster modifiers, returned by the character API as "tamedBeastProperties". These were previously discarded on import, so companions were modeled only as their base monster.

- Import: parse tamedBeastProperties into a per-gem modifier list, resolving mod ids from "[ModId |Display]" tokens with a display-line fallback for nameplate-only lines; user enable/disable choices survive re-import
- Data: new generated Data/TamedBeastMods.lua (export script iterates the Monster-domain mod pool, naming entries from KeywordPopups/ArchnemesisMods and resolving effects through SkillStatMap); mods with a positive spawn weight are flagged rollable and offered in the UI
- Calcs: enabled modifiers apply to the companion minion's modDB, gated on the active skill being a Companion so stale lists never leak
- UI: "Beast Modifiers" section in the socket group editor with per-row selection dropdown, enable checkbox, remove button and calc-comparison tooltips, mirroring the gem slot layout
- Persistence: modifiers saved as TamedBeastMod child elements of Gem; old builds are unaffected

image


src/Modules/CalcPerform.lua
```diff
@@ -1026,7 +1026,8 @@ function calcs.perform(env, skipEHP)
 		if mainSrcInstance and mainSrcInstance.tamedBeastModList and mainGrantedEffect and mainGrantedEffect.minionList and mainGrantedEffect.name:match("^Companion") then
 			for _, beastMod in ipairs(mainSrcInstance.tamedBeastModList) do
 				local beastModData = beastMod.enabled and beastMod.modId and env.data.tamedBeastMods[beastMod.modId]
-				if beastModData then
+				-- A mod the beast's tags can never roll is ignoreed
```
Nightblade, line 1029: Typo:  "ignoreed" should be "ignored".

PR #2159, opened 2026-06-11
Add support for Loyalty/damage taken from Companion's Life before you

ltogniolli:
Implements the "% of Damage from Hits is taken from your (Damageable) Companion's Life before you" mechanic as a new "companion" ally pool in the existing alliesTakenBeforeYou framework, covering:

- Tree nodes, Starkonja's Head, and the Bonded rune (hit redirect)
- Forgotten Warden (Deflected Hits variant, scaled by Deflect chance)
- Loyalty and Romira's Requital support gems via a shared statMap entry (Romira's recoup is left unsupported for a follow-up)

Total companion life is summed automatically from all active damageable companion skills: the main-skill minion init in CalcPerform is extracted into initMinionModDB so non-main companions can be initialised the same way, with a "Total Companion Life" config field as a manual override.

Also drops the stale ifSkill="Raise Spectre" gate from the Total Spectre Life config field so it is reachable in PoE2, and regenerates the four mis-parsed companion entries in ModCache.

Manual/unit testing.

https://pobb.in/70HXySRg9wV5

### Before screenshot:
image
image

### After screenshot:
image
image
image


LocalIdentity (approved):
I made a few fixes and improvements
It now works with the spectre mod too
Fixed Loyalty not stacking correctly when used on multiple companions
Changed the config option to state override as that's all its needed for now

PR #2160, opened 2026-06-11
Fix Full DPS breakdown and gem dropdowns showing the wrong name for Companion / Spectre skills

ltogniolli:
Companion and Spectre gems share a single granted effect ("Companion: {0}" / "Spectre: {0}") whose name is mutated globally for reservation display, so every entry in the Full DPS breakdown showed the same minion name. Derive each entry's name from the skill's own minion instead.

https://pobb.in/-TeCx6SJiuWv

### Before screenshot:
image

### After screenshot:
image

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -377,6 +377,27 @@ describe("TestSkills", function()
 		end
 	end)
 
+	it("uses selected companion names in skill displays", function()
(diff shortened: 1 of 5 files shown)


LocalIdentity (approved):
I moved the function to calcs so that we can also use it to fix the same isssue with the wrong name being used for the Active Skill dropdown

PR #2161, opened 2026-06-11
Add missing tamable beasts and their skills

ltogniolli:
Add 26 beast spectre/companion varieties:
- Anundr, the Sandworm
- The Ravenous Fang (map variant)
- Rasp Scavenger (medium, large and Sekhema Trial sizes)
- Black Strider (incl. Sekhema Trial variant)
- Porcupine Crab
- Venomous Crab Matriarch
- Tarnished Beetle
- Adorned Scarab
- Adorned Beetle
- Hyena Demon
- Sun Clan Scavenger
- Tombshrieker (Sekhema Trial variant)
- Plague Nymph
- Nymph Wasp
- Chaw Mongrel
- Moray Clan
- Armoured Parasite
- Kreth Parasite
- Feral Primate
- Flesh Larva (ranged variant)
- Tendril Prowler
- Rabid Dog

Add the monster skills these beasts use to the spectre skill export; this also fills in the previously-empty kits shared by The Devourer, The Ravenous Fang and Rasp Scavenger.

The spectre list report now counts same-name monsters with different Spirit reservations as distinct variants instead of duplicates, and beast-category monsters are no longer excluded by the boss-health-bar filter, since boss beast companions like Anundr carry it.

### DIFF
--- a/src/Data/Skills/spectre.lua
+++ b/src/Data/Skills/spectre.lua
@@ -8281,3 +8281,2281 @@ skills["GAIcyQuadrillaBossRectSlam"] = {
 		},
 	}
 }
+
+skills["MeleeMudBurrowerLeftCleave"] = {
+	name = "Left Cleave",
+	hidden = true,
+	icon = "Art/2DArt/SkillIcons/iconbasicattack.dds",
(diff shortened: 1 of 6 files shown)


LocalIdentity (approved):
We now have 2 Rasp Scavenger and Black Strider entries so we can't be sure that the correct one will be used on import but if people do use either one then its probably good to have the option

PR #2184, opened 2026-06-12
Performance improvements for gem dropdown sort

ltogniolli:
Opening the gem dropdown can take a long time. This is specially noticeable when sorting by Full DPS when you have several sources of full dps enabled. Two sets of improvements:

The first commit adds a guard to avoid recalculating the support skill tooltip multiple times per second on hover. It also implements the fastCalc path partially introduced on #1504 that avoids recalculating unnecessary things when in Full DPS mode. This gives a modest 10% improvement on my testing.

The second commit adds a more aggressive cache that avoids getting into the calculation at all by storing inputs and outputs of the previous calculation only recalculating Full DPS targets when the change input impact that skill. It provides a much better 60-70% improvement on my testing.

This cache only applies to the fastCalc/FullDPS path so existing callers are unnaffected decreasing the risk.

It doesn't include my changes from #2147 . if that lands first the caching needs to be updated, and vice versa.

I automated runs in a few of my builds making sure the cached/uncached values match, created an automated test spec with the main assertions and edge cases encountered, and manual testing.

### Before screenshot:

image

### After screenshot:

image


LocalIdentity (approved):
Fixed an issue with triggered skills but I think this looks good
Does merge a bunch of the previous logic into loops so it makes it easier to expand in the future too

PR #2190, opened 2026-06-13
Fix not being able to highlight text in compare config tab

vaisest:

Fixes compare tab config subtab controls clearing user selection on every frame.

~~Also fixes a crash when distance config has too many digits that I accidentally found~~ Reverted because @LocalIdentity had a fix implemented elsewhere

- Tested with my own build while using pob

https://pobb.in/x5Djmznj897s
https://pobb.in/CZfC5voLQXWp
### Before screenshot:
https://github.com/user-attachments/assets/f9b99948-7c29-46d5-909f-7c6625851728
### After screenshot:
https://github.com/user-attachments/assets/9db60c01-236e-4ef1-8caf-5fd46054be8b


LocalIdentity (approved):
I fixed it not allowing you to type in `-` anymore with the focus part

PR #2208, opened 2026-06-15
Add item skill tooltips

vaisest:

Adds tooltips for item skills. Uses the level defined on the item. Displayitem not included, as it's not really a tooltip and would be weird to implement. Tooltips show when holding shift and are attached to the right of the item tooltip, and grow downwards. They will move over the main tooltip, which should be fine as the idea is that you can explicitly look at the skill, instead of it being e.g. shown at all times.

- Just about every DB item tested
- Real build items tested

### After screenshot:

https://github.com/user-attachments/assets/f5b4d313-ff45-488f-8186-2d1e90584294


vaisest:
I think this worked pretty well, and I couldn't find any remaining issues, but I wasn't sure if this was preferred over e.g. always showing it and having smarter tooltip logic

LocalIdentity (approved):

Issue #2217, opened 2026-06-17
Trying to import Skills from Disciple of Varashta Crashes and Bricks POB


LonleyTurtle:
### Check version

- [x] I'm running the latest version of Path of Building and I've verified this by checking the [changelog](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/blob/master/CHANGELOG.md)

### Check for duplicates

- [x] I've checked for duplicate open **and closed** issues by using the search function of the [issue tracker](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2/issues?q=is%3Aissue)

### What platform are you running Path of Building on?

Windows

### Screenshots of crash/graphical issue, version and options menu

I am trying to import my Build into POB, have imported several builds over the course of the year without issue. POB is always updated. Recently I started a disciple of Varashta Build and importing items and skills works fine. However any time I import passive tree and jewels it crashes and bricks with this error. The first time I ported in the passives/abilities the issue did not occur until I hovered over an ability in the POB menu. This then crashed and since the profile was saved, it would always open up to the error below. I had to do a complete wipe and DL the Portable version losing all my prior saved builds. 

Error:

In 'OnFrame': Modules/Build.lua:2044: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
	[C]: in function 'ipairs'
	Modules/Build.lua:2044: in function 'RefreshSkillSelectControls'
	Modules/Build.lua:1344: in function 'CallMode'
	Modules/Main.lua:372: in function <Modules/Main.lua:342>
	[C]: in function 'PCall'
	Launch.lua:111: in function <Launch.lua:108>
v0.21.0 master
Press Enter/Escape to dismiss, or F5 to restart the application.
Press CTRL + C to copy error text.

### What are your system specifications and configuration?

OS Name	Microsoft Windows 11 Home
Processor	AMD Ryzen 9 5900X 12-Core Processor, 4176 Mhz, 12 Core(s), 24 Logical Processor(s)
BaseBoard Manufacturer	Micro-Star International Co., Ltd.
BaseBoard Product	MEG X570S ACE MAX (MS-7D50)
Video Card Nvidia GeForce 3080TI  Founders edition

### How to reproduce the issue

Click import passives and abilities

### PoB for PoE2 build code

```shell
https://poe.ninja/poe2/profile/Kosmo2409-4451/runesofaldur/character/KozzieCull
```

[...]

PR #2230, opened 2026-06-18
Add option to return to build selection when crashing

vaisest:
Fixes issues where users encounter an issue which crashes PoB when loading. This is obviously not very common, but in cases such as #2217 can cause PoB to crash when loading, which renders both the dismiss and restart options useless. There can also be crashes in Draw functions, such as when it happened to timeless jewels at the start of 0.5 due to missing textures.

This option lets users at least continue using PoB without having to manually remove the problematic profile file, which they probably won't know to do.
- Yes
- Can be tested by e.g. adding an `error()` to a tab's Draw method

https://pobb.in/wS-S5R5EYZfC (#2217 build which crashes)
### Before screenshot:
image

### After screenshot:
image

### DIFF
--- a/src/Launch.lua
+++ b/src/Launch.lua
@@ -357,6 +357,9 @@ function launch:ShowPrompt(r, g, b, str, func)
 	self.promptFunc = func or function(key)
 		if key == "RETURN" or key == "ESCAPE" then
 			return true
+		elseif key == "F4" then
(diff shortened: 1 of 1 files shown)


LocalIdentity (approved):
Fixed a small issue when a crash occurred inside a popup, but this is a good fix to have for the future

PR #2246, opened 2026-06-19
Add support for "+X% to Fire Spell Critical Hit Chance" mod

### LINKED ISSUE #2226: support for "+X% to Fire Spell Critical Hit Chance" affix

luther-rotmg:
Closes #2226.

### Problem
The "of Xoph" suffix — `+X% to Fire Spell Critical Hit Chance` (mod group `FireSpellBaseCriticalChance`) — had no `modNameList` entry, so the `fire spell` qualifier was left unparsed and the affix granted nothing.

### Fix
Adds a `modNameList` mapping to `CritChance` (BASE) restricted to fire spells via `ModFlag.Spell` + `KeywordFlag.Fire`, matching the existing `fire spells` convention already used in `preFlagList`.

### Testing
Adds a parse regression test in `spec/System/TestItemParse_spec.lua`: `+5% to Fire Spell Critical Hit Chance` grants +5 base crit to fire spells, and 0 to attacks / 0 to cold spells. The full test suite and the ModCache check pass.

### DIFF
--- a/spec/System/TestItemParse_spec.lua
+++ b/spec/System/TestItemParse_spec.lua
@@ -73,6 +73,21 @@ describe("TestItemParse", function()
 		assert.are.equals(12, item.quality)
 	end)
 
+	it("parses '+X% to Fire Spell Critical Hit Chance' (issue #2226)", function()
+		local item = new("Item", [[
+			Rarity: Rare
+			Xoph's Test Band
+			Amethyst Ring
+			Implicits: 0
+			+5% to Fire Spell Critical Hit Chance
+		]])
+		-- grants base critical hit chance to fire spells specifically
+		assert.are.equals(5, item.baseModList:Sum("BASE", { flags = ModFlag.Spell, keywordFlags = KeywordFlag.Fire }, "CritChance"))
+		-- must NOT apply to attacks, nor to non-fire spells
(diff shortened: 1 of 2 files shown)


Blitz54:
This should be done differently. It should be under ```-- Skill Types``` as just ```fire spells``` so that it works for other mods like crit damage bonus. Worth adding similar lines for cold, lightning, and chaos spell for future proofing.

luther-rotmg:
> This should be done differently. It should be under `-- Skill Types` as just `fire spells` so that it works for other mods like crit damage bonus. Worth adding similar lines for cold, lightning, and chaos spell for future proofing.

Gotcha, give me a chance to rework it.

luther-rotmg:
Thanks for the review — agreed, that's a cleaner approach. Pushed an update:

- Dropped the single `modNameList` entry and instead added `fire` / `cold` / `lightning` / `chaos spell` to `modFlagList`'s `-- Skill types` section. Since `modFlagList` is scanned after the mod name, the element+spell qualifier now composes with **any** stat through the normal parse — so it handles `Critical Damage Bonus` as well as `Critical Hit Chance` — and the other elements are future-proofed.
- Updated the regression test to assert the tag composes with both `CritChance` and `CritMultiplier`, and works per-element (cold).

Tests + ModCache check are green.

LocalIdentity (approved):

PR #2250, opened 2026-06-20
Support Vestige of Darkness physical damage to Chill/Freeze modifier

bct8925:
## Description

Adds support for the **Vestige of Darkness** (Tenebrous Crown) unique helmet affix:

> **Physical damage from Hits Contributes to Chill Magnitude and Freeze Buildup**

This modifier was previously unrecognised by the mod parser, so equipping the helmet had no effect on Chill or Freeze.

## Changes

- **`src/Modules/ModParser.lua`** — new rule emitting `flag("PhysicalCanChill")`, `flag("PhysicalCanFreeze")`.
- **`src/Data/ModCache.lua`** — regenerated the stale cache entry for this line (it was cached as unrecognised, which shadowed the new parser rule).
- **`spec/System/TestAilments_spec.lua`** — new system test: a pure-physical Quarterstaff Strike has no Freeze buildup baseline, then gains Freeze buildup and Chill once the modifier is applied.

## Testing

- Full Busted suite passes (429 successes / 0 failures).
- Verified end-to-end with the actual unique equipped (ModCache loaded, normal app path): with the helmet, physical hits drive `FreezeBuildupAvg > 0` and apply Chill; without it both remain zero.

## Reference build

Example build affected by this modifier: https://maxroll.gg/poe2/pob/4a65hh0h

### DIFF
--- a/spec/System/TestAilments_spec.lua
+++ b/spec/System/TestAilments_spec.lua
@@ -28,33 +28,4 @@ describe("TestAilments", function()
 
 		assert.are.equals(1.1, build.calcsTab.mainOutput.PoisonEffMult)
 	end)
-
-	it("physical damage contributes to chill and freeze with Vestige of Darkness modifier", function()
-		-- High flat physical, no cold, so any chill/freeze must come from physical
-		build.itemsTab:CreateDisplayItemFromRaw([[
-			New Item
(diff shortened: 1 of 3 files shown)


LocalIdentity (approved):
Removed test as it's not needed

PR #2251, opened 2026-06-20
Fix Demon Form not counting as being Shapeshifted

cupkax:
Fixes # .

Turning on demon form in config did not apply the shape shifted condition.

Verified passive tree shapeshifted nodes to check whether the condition was being verified true:
- Life Regen on shapeshift
- Max Lightning res on shapeshift

https://poe.ninja/poe2/pob/20fb0

### DIFF
--- a/src/Modules/CalcPerform.lua
+++ b/src/Modules/CalcPerform.lua
@@ -369,6 +369,9 @@ local function doActorAttribsConditions(env, actor)
 			modDB:NewMod("StunThreshold", "INC", 50, "Wyvern Form")
 			modDB:NewMod("AilmentThreshold", "INC", 50, "Wyvern Form")
 		end
+		if env.configInput.inDemonForm then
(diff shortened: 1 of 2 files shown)


cupkax:
This is a very hacky fix as I just read the check box and make sure shapeshifted returns true. Prob a better a way to do this.

I tried checking through modb flag but I think the minimum life calculations messed with it a tad

Blitz54:
This could just be done in the ConfigOptions by adding the shapeshifted mod to the Demon Form config.
```modList:NewMod("Condition:Shapeshifted", "FLAG", true, "Config")```

cupkax:
> This could just be done in the ConfigOptions by adding the shapeshifted mod to the Demon Form config. `modList:NewMod("Condition:Shapeshifted", "FLAG", true, "Config")`

Done, ta

LocalIdentity (approved):
Removed the CalcPerform line

PR #2255, opened 2026-06-21
Fix number of times block recently gated behind unique enemies

### LINKED ISSUE #2254: Movement speed / block parsed but not functional
### Check version

### Check for duplicates

Windows

The modifier `1% increased movement speed for each time you've blocked in the past 10 seconds` (from the passive skill tree `Swift Blocking`) should enable something like `Have you blocked in the past 10 seconds? / # of times`.

This mod seems to be parsed correctly but is not functional. Once you allocate the notable (or add as a custom mod), nothing changes in the configuration tab.

_No response_

### PoB for PoE2 build code

```shell

```

### Screenshots

Parsed:
Image

But not functional:
Image

cupkax:
Fixes #2254.

Number of times blocked recently condition was gated behind unique enemies. Generalised it to make sure the condition returns true regardless of block context.

- `Swift Blocking` notable works to give movespeed based on input value in config


Blitz54:
Since we're editing this line anyway, maybe change the var name to ```multiplierBlockedPast10Sec``` just to match the majority of the other configs. Thanks for fix!

cupkax:
> Since we're editing this line anyway, maybe change the var name to `multiplierBlockedPast10Sec` just to match the majority of the other configs. Thanks for fix!

Thanks, updated the pr

LocalIdentity (approved):

PR #2277, opened 2026-06-23
Add support for Sinister Jewel sockets

followingthefasciaplane:
## Description: 

intended to add support for sinister jewel sockets granted by Voices and Zarokh's Gift. granted sockets now apply in calculator environments, update main build socket state only during the main calc pass, and are removed cleanly when the granting anoint/item goes away.

also wanted to fix the anoint picker so Zarokh's Gift appears by adding the new 0.5 emotions to the filter list, and verify tooltip comparisons include jewels socketed.

dynamically adds or removes socketables from the items UI with persistent socket allocation

- test spec can be found in commit
- preliminary testing allocating and unallocating sinister sockets
- needs a more experienced contributor to review, im not entirely confident 

### Screenshots

z6
z5
z4
z3
z2
z1


civilianlp:
great work; thank you. However, would there be any value to explicitly name the sockets as "Sinister Sockets #<n>" for visual clarity? This way you can also block uniques in dropdown? (unless you're already doing so)?

PR #2283, opened 2026-06-24
Fix some Bonded mods not working

Blitz54:
Some basic Bonded mods did not parse properly. Mods like `Bonded: Minions have 30% increased Area of Effect` failed because the Bonded: part would match the preFlag, so then the Minion part never matched.

This was done by AI mostly as I didn't feel like braining. Will look at it another day, setting as draft for now.

### Before screenshot:
image

### After screenshot:
image


LocalIdentity:
Superseded by #2366

PR #2290, opened 2026-06-24
Add required stats filter to trader

### LINKED ISSUE #471: Trader "Required Stats" option.
### Check for duplicates

Windows

Using the trader mechanic, I'd like to add an additional "Required Attributes" section as part of the weighting process. For example, my build may have "Spirit" requirements and without it, my build would be bricked. 

I'd like to be able to add required attributes to my market searches. For example, Spirit, Additional Projectiles, etc... this would ensure that the results I get back, can be added to my current build, and it's the best for my money as far as DPS and EHP is concerned. 

I've been forced to manually add into the trade website min & max on some attributes just to ensure that I'm not bricking my build. I've also had to add stats, however that rarely works in conjunction with weights. 

### Additional context

_No response_

vaisest:
Fixes #471.

Adds a UI which can be used to add an AND filter to the query. There are quite a few old issues about this in the pob1 repo. This is especially useful for e.g. spirit, filling attributes, or corrupted blood jewels in poe1

This also fixes an issue where a row didn't get cleared when results were empty.

- All slots tested
- Many modifiers tested

### Before screenshot:
image

### After screenshot:

https://github.com/user-attachments/assets/6962d71f-ace8-4c15-8ba7-3ce9fcf97e12

image

image

### DIFF
--- a/src/Classes/CompareBuySimilar.lua
+++ b/src/Classes/CompareBuySimilar.lua
@@ -347,19 +347,7 @@ function M.openPopup(item, slotName, primaryBuild)
 		uri = result
 	end
 
-	-- Helper: create a numeric EditControl without +/- spinner buttons, and
-	-- with a preset changeFunc
-	local function newPlainNumericEdit(anchor, rect, init, prompt, limit, integer)
-		local format = integer and "%D" or "^%d."
-		local ctrl = new("EditControl", anchor, rect, init, prompt, format, limit, rebuildUrl)
-		-- Remove the +/- spinner buttons that "%D" filter triggers
-		ctrl.isNumeric = false
-		if ctrl.controls then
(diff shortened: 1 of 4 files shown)


vaisest:
I considered adding a NOT section that would blacklist some mods from having weights generated, but I can't really see this being comfortable to use. There are very few things that actually need it in rare gear, and I think it's probably mostly just cull which we might as well manually blacklist.

Megalomaniac have mods which inflate values for sure, but it would be quite uncomfortable to use. Manually editing the search and using price item is probably faster than doing the whole thing again, because it's likely you'll only realise it's giving you bad notables when you look at the results. It would also be more difficult to make a searchable list of the nodes.

LocalIdentity (approved):

PR #2292, opened 2026-06-24
Fix gem group counting logic

moxaj:

Before, when you had a granted skill from an item / tree, and disabled that socket group, it would be counted towards the limit of 9.

- equip 9 `Arc` gems, each in its own group
- equip `Sadist's Mercy Flanged Mace`
- disable the granted `Harbinger of Madness`
- verify that there is no warning message for going over the limit of 9
- add another socket group with `Arc` and `Ice Nova` (in the same group)
- verify that there's a warning for the limit iff the socket group is enabled and at least one of `Arc` and `Ice Nova` is enabled

### DIFF
--- a/spec/System/TestSkills_spec.lua
+++ b/spec/System/TestSkills_spec.lua
@@ -539,6 +539,8 @@ describe("TestSkills", function()
 
 		local skillsTab = {
 			socketGroupList = {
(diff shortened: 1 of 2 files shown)


LocalIdentity (approved):
I simplified the fix

PR #2307, opened 2026-06-26
Remove `Are you fortified?` config option

moxaj:

Fortify (buff, support gem, fortification stacks) does not exist in poe2. This PR removes the `Are you fortified?` configuration option.

**Question**: shall I remove everything related to fortify, or keep it in case they add it to the game in the future?


cupkax:
Duelist would prob add some more auras. Fortify might be one of em. Might be better to wait until 1.0 for this?

LocalIdentity (changes requested):
Better to just hide the config instead of removing it

moxaj:
@LocalIdentity by hiding you mean just commenting out the config instead of deleting it?

LocalIdentity (approved):

PR #2308, opened 2026-06-26
Fix formatting of 'Waistgate Heavy Belt' entry

Pcodcsja:
The name of this unique belt contains repeated content

### DIFF
--- a/src/Data/Uniques/belt.lua
+++ b/src/Data/Uniques/belt.lua
@@ -267,7 +267,7 @@ Your Life Flask also applies to your Minions
 Minions cannot Die while affected by a Life Flask
 (20-30)% increased Flask Charges gained
 ]],[[
-Waistgate Heavy Belt
+Waistgate
 Heavy Belt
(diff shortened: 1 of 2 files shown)


Blitz54:
Nice catch, but we'll need the fix in `src/export/uniques/belt.lua` too otherwise when we export uniques, it will come back. Guess we should add a comment to the top of the unique files that mention they are generated.

LocalIdentity (approved):
All the uniques are generated from the respective txt file so you are meant to make changes there and then run the export script to make the change to the corresponding lua file fyi

Pcodcsja:
> All the uniques are generated from the respective txt file so you are meant to make changes there and then run the export script to make the change to the corresponding lua file fyi所有暗金装备都是从相应的 txt 文件生成的，因此你应该在该文件中进行修改，然后运行导出脚本，将更改同步到对应的 lua 文件中，特此说明。

Thank you for your guidance. This is my first time working on this project, and I will pay more attention next time!

PR #2312, opened 2026-06-27
Add support for Atziri's Splendour mana to life cost conversion for corrupted gems

slaren:
The mod "Skills from Corrupted Gems have 50% of Mana Costs Converted to Life Costs" from Atziri's Splendour is not supported.

- Toggle between corrupted and uncorrupted skill gem
- Verify that skill costs match the game tooltip

I am not sure if this is the best way to add the "GemCorrupted" condition, any feedback is appreciated.

### DIFF
--- a/src/Modules/BuildDisplayStats.lua
+++ b/src/Modules/BuildDisplayStats.lua
@@ -77,7 +77,7 @@ local displayStats = {
 	{ stat = "AreaOfEffectRadiusMetres", label = "AoE Radius", fmt = ".1fm" },
 	{ stat = "BrandAttachmentRangeMetre", label = "Attachment Range", fmt = ".1fm", flag = "brand" },
 	{ stat = "BrandTicks", label = "Activations per Brand", fmt = "d", flag = "brand" },
-	{ stat = "ManaCost", label = "Mana Cost", fmt = ".0f", color = colorCodes.MANA, pool = "ManaUnreserved", compPercent = true, lowerIsBetter = true, condFunc = function(v,o) return o.ManaHasCost end },
+	{ stat = "ManaCost", label = "Mana Cost", fmt = "d", color = colorCodes.MANA, pool = "ManaUnreserved", compPercent = true, lowerIsBetter = true, condFunc = function(v,o) return o.ManaHasCost end },
 	{ stat = "ManaPercentCost", label = "Mana Cost", fmt = "d%%", color = colorCodes.MANA, pool = "ManaUnreservedPercent", compPercent = true, lowerIsBetter = true, condFunc = function(v,o) return o.ManaPercentHasCost end },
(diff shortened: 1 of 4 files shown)


src/Modules/BuildDisplayStats.lua
```diff
@@ -77,7 +77,7 @@ local displayStats = {
 	{ stat = "AreaOfEffectRadiusMetres", label = "AoE Radius", fmt = ".1fm" },
 	{ stat = "BrandAttachmentRangeMetre", label = "Attachment Range", fmt = ".1fm", flag = "brand" },
 	{ stat = "BrandTicks", label = "Activations per Brand", fmt = "d", flag = "brand" },
-	{ stat = "ManaCost", label = "Mana Cost", fmt = "d", color = colorCodes.MANA, pool = "ManaUnreserved", compPercent = true, lowerIsBetter = true, condFunc = function(v,o) return o.ManaHasCost end },
+	{ stat = "ManaCost", label = "Mana Cost", fmt = ".0f", color = colorCodes.MANA, pool = "ManaUnreserved", compPercent = true, lowerIsBetter = true, condFunc = function(v,o) return o.ManaHasCost end },
```
slaren: This is somewhat unrelated, but while working on this I noticed that the mana cost display was not being rounded, so in some cases it would be off by one when compared to the in-game tooltip.

src/Modules/CalcActiveSkill.lua
```diff
@@ -694,6 +694,9 @@ function calcs.buildActiveSkillModList(env, activeSkill)
 		activeSkill.weapon2Cfg.skillCond = setmetatable({ ["OffHandAttack"] = true }, { __index = activeSkill.skillCfg.skillCond })
 		activeSkill.weapon2Cfg.flags = bor(skillModFlags, activeSkill.weapon2Flags)
 	end
+	if activeEffect.srcInstance and activeEffect.srcInstance.corrupted then
+		activeSkill.skillCfg.skillCond["GemCorrupted"] = true
+	end
```
Blitz54: Could this be moved to line 771? As it's already checking for corrupted there.

```
	if activeSkill.activeEffect.srcInstance and activeSkill.activeEffect.srcInstance.corrupted and not (activeSkill.activeEffect.srcInstance.fromItem or activeSkill.activeEffect.srcInstance.fromTree or activeSkill.activeEffect.grantedEffect.fromItem or activeSkill.activeEffect.grantedEffect.fromTree) then
		skillModList:NewMod("GemCorruptionLevel", "BASE", activeSkill.activeEffect.srcInstance.corruptLevel, "Corruption")
	end
```

slaren: Done.

LocalIdentity (approved):
We don't round the display of mana costs, they shouldn't be showing up as decimal values and should be handled at the calculation point in CalcOffence. If you have an example with incorrect rounding, then feel free to make an issue, and we can look into it

PR #2323, opened 2026-06-29
Fix level requirements on staves, wands, and jewellery and unique items

Blitz54:
0.5.4 changed granted skills, so that the item does not require level or attributes for the granted skill. The requirements are either from the base item, or specifically for the unique (Sylvan's Effigy level 62 on a Stoic Sceptre which is level 6).

- Advanced copy paste format is different from poe1, so some changes were made to parse the "Level Requirement". Int, Str, and Dex are obtained from the base item and calculated through mods later, so those have been stripped.
- Jewellery level req works like before I broke it. It uses the modLine Level, and scales that by 0.8 for the level req.
- Import isn't 100%. I left a comment where things should be fixed at some point. But it's a little involved atm.
- Wands, sceptres, and staves have a base level req for crafting now.


LocalIdentity (approved):
Will need to update that code once we move to using mod IDs as it will simplify all this a bunch

PR #2339, opened 2026-06-30
Fix buy similar popup not updating url and persist selections

vaisest:

Fixes the edit boxes and realm/league/trade type selectors not changing the resulting trade url. They were included in the result, but changing them didn't update the url so only toggling the checkbox actually applied the value

Thanks @moxaj for reporting this

946d81e also fixes a crash when the socket you're viewing doesn't exist on your tree. Probably not relevant currently, but missing cluster jewel sockets were causing crashes in PoB1 and it makes sense here.

It also makes the upper selector choices remembered when reopening the popup

-


LocalIdentity (approved):
I fixed an issue with changing the dropdowns not updating the URL so you probs want to port that to your PoB 1 PR

vaisest:
> I fixed an issue with changing the dropdowns not updating the URL so you probs want to port that to your PoB 1 PR

I had already fixed this in the PoB 1 PR but I seem to have forgotten to add it here too. Oops. I'll copy your fix for parity's sake

PR #2346, opened 2026-07-03
Fix "Shapeshifted" condition being toggled on every build.

briansonnenberg:
Fixes "Shapeshifted" condition being toggled on every build.

The shapeshifted condition is always true on every build.

- Checked that the shapeshifter nodes don't show bonuses unless:
1. the "Currently Shapeshifted" config is toggled
2. Demon Form is an enabled skill gem (new behavior)
3. Any of the shapeshift gems are the main skill gem (old behavior)

Any build.

### Before screenshot:

Screenshot 2026-07-03 000932

### After screenshot:

Screenshot 2026-07-03 081905

### DIFF
--- a/src/Data/Skills/other.lua
+++ b/src/Data/Skills/other.lua
@@ -4458,6 +4458,9 @@ skills["DemonFormPlayer"] = {
 			},
 			baseFlags = {
 			},
+			baseMods = {
+				flag("Condition:Shapeshifted", { type = "GlobalEffect", effectType = "Buff" }, { type = "Condition", var = "DemonForm" }),
+			},
 			constantStats = {
 				{ "demon_form_spell_damage_+%_final_per_stack", 3 },
 			},
--- a/src/Export/Skills/other.txt
+++ b/src/Export/Skills/other.txt
@@ -304,6 +304,7 @@ statMap = {
 		mod("LifeDegen", "BASE", nil, 0, 0, { type = "Condition", var = "DemonForm" }, { type = "GlobalEffect", effectType = "Buff", effectName = "Demon Form" }, { type = "Multiplier", var = "DemonFlameStacks", limitVar = "DemonFlameMaximum" } ),
 		div = 60,
 	},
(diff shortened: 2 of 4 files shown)


majochem:
I think this could also be fixed, by just removing the `defaultState = true` from the demon form config? 
That is why it auto toggles to true, even if the option is not available. Unless there's a good reason why it should default to true

```
{ var = "inDemonForm", type = "check", label = "Are you in Demon Form?", ifSkill = "Demon Form", defaultState = true, tooltip = "Players need a minimum of 2 ^xE05030Life ^7to enter Demon Form, so you cannot use it with Chaos Inoculation", apply = function(val, modList, enemyModList)
		modList:NewMod("Condition:Shapeshifted", "FLAG", true, "Config")
		modList:NewMod("Condition:DemonForm", "FLAG", true, "Config", { type = "StatThreshold", stat = "Life", threshold = 2 })
end },
```

briansonnenberg:
> I think this could also be fixed, by just removing the `defaultState = true` from the demon form config? That is why it auto toggles to true, even if the option is not available. Unless there's a good reason why it should default to true
> 
> ```
> { var = "inDemonForm", type = "check", label = "Are you in Demon Form?", ifSkill = "Demon Form", defaultState = true, tooltip = "Players need a minimum of 2 ^xE05030Life ^7to enter Demon Form, so you cannot use it with Chaos Inoculation", apply = function(val, modList, enemyModList)
> 		modList:NewMod("Condition:Shapeshifted", "FLAG", true, "Config")
> 		modList:NewMod("Condition:DemonForm", "FLAG", true, "Config", { type = "StatThreshold", stat = "Life", threshold = 2 })
> end },
> ```

The reason that this defaults to true is because the intent is "If Demon Form is an active gem, then default inDemonForm to true".  That makes sense.  If the build uses DemonForm it's probably safe to assume that they want to calculate DemonForm damage, not have to go toggle it on via the skill configuration manually.  The problem appears to be that the apply of the ModList takes effect regardless of the skill gem check, so that's why I added the gating there.  Come to think of it, I guess I need to add the gating based on the HP to the shapeshift condition as well.

LocalIdentity (approved):
I simplified it by moving the condition to the Demon Form gem

PR #2359, opened 2026-07-08
Fix Heavy Stunned enemies not counting as Stunned

moxaj:

If an enemy is heavy stunned, they should also be considered stunned (duh).

- Add these custom configurations:
```
Damage with hits is lucky against heavy stunned enemies
+25% to critical damage bonus against stunned enemies
```
- Add `Arc`
- Tick `Is the enemy Heavy stunned?`
- Verify that the crit multi of `Arc` is `225%`.


majochem:
> +25% to critical damage bonus against stunned enemies

I guess it doesn't hurt, but is this wording actually still present on a passive skill or item in game? I thought PoE2 now uses "Heavy Stunned" for all of those conditions.

moxaj:
@majochem The passive skill tree notable `Sundering` uses this wording.

majochem:
> @majochem The passive skill tree notable `Sundering` uses this wording.

Interesting. Thank you for confirming. Does indeed seem useful then, even if I think the wording on that node is likely to be adjusted by GGG in the future.

LocalIdentity (approved):

PR #2378, opened 2026-07-14
Fix mod parsing for "enemies you <verb> have ..."

moxaj:

Some mods like `Enemies you Curse have -6% to Chaos Resistance` were parsed but did not function.

This PR:
- fixes the generic `enemies you .. have ..`
- fixes `enemies you .. cannot recharge energy shield`
- removes the redundant `enemies you electrocute have ..` and `enemies you mark have #% reduced accuracy rating`

- Add `Essence Drain`
- Add `Temporal Chains`
- Add the custom configuration `Enemies you Curse have -6% to Chaos Resistance`
- Verify that your damage increases


src/Modules/ModParser.lua
```diff
@@ -25,6 +25,17 @@ local function combineToUpper(str, sepPattern)
 	end
 	return outStr
 end
+
+-- Convert condition verb forms to their proper condition variable names
+local function getConditionVar(verb)
+	local conditionMap = {
```
majochem, line 32: Maybe this `conditionMap` could go next to `statusToEffectMap` or otherwise be combined in a clever way, as they are essentially just converting the same terms in opposite directions.

moxaj: I can see the connection, but I'm not sure how they could be combined. There's not much of an overlap between the two.

majochem: yeah, not sure "combining" is actually feasible, but it would be nice to only maintain one if someone actually figured out a way. 

My main point was that imo they should be located in similar places in the code though since they serve similar functions. It's just that mod wording is not 100% consistent for e.g. "Enemies you ..." and "Enemies ... by you" and similar mods. Some use the verb, some the participle, which is why we apparently need to map both ways.

No idea which place is actually better in the code, but having those maps close to each other does seem to make sense 🤷

moxaj: added some notes regarding naming / combining

PR #2387, opened 2026-07-21
Add legacy flag to Commandment support

moxaj:

Commandment was missing the legacy flag, this PR fixes that.

- Verify that commandment is only available if you enable `Show legacy gems`


LocalIdentity (approved):
All the gem lua files are generated from the corresponding txt file. Without adding it to the txt file the legacy flag would be removed the next time I exported the gems

Issue #2415, opened 2026-08-15
Cadiro's Gambit: Perandus Arrows are not supported


lucaskarsten:
### Check for duplicates

I've searched open and closed issues for "Cadiro" and "Perandus" and found nothing.

### What platform are you running Path of Building on?

Windows

### Is your feature request related to a problem?

Cadiro's Gambit (Primed Quiver) has the line "Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow", which is unsupported. The quiver currently calculates as if it only granted increased Attack Speed.

The Diamond Arrow is the one that matters for the DPS number: all its Hits are Critical Hits, with 60% increased Critical Damage Bonus. On a bow character with 29% Critical Hit Chance and a 4.28x Critical Damage Bonus, that Arrow alone is worth 903k Full DPS against 393k, so the item reads as far weaker than it is.

### Describe the solution you'd like

The six Perandus Arrows are:

- Crescendo: Chain 6 additional times, 60% increased Damage for each time they have Chained
- Splinter: Split towards 6 targets
- Reversing: Pierce all targets and Return to you
- Diamond: all Hits are Critical Hits with 60% increased Critical Damage Bonus
- Covetous: enemies killed drop items with 600% increased Rarity
- Blunt: 600% increased Stun Buildup

Each Arrow is rolled independently, so averaging behaviours as different as Chaining and guaranteed Critical Hits into a single number is not meaningful. A config dropdown that picks which Arrow to calculate, in the shape of the existing "Random element mode" for modifiers that choose an element, keeps every number honest and leaves the default behaviour unchanged.

### Describe alternatives you've considered

Applying all six at once, which would report a character that Chains, Splits, Pierces and always Crits simultaneously. That number does not exist in game.

PR #2416, opened 2026-08-15
Support Cadiro's Gambit's Perandus Arrows

lucaskarsten:
Related to #2415.

Cadiro's Gambit has the line "Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow", which was unsupported, so the quiver calculated as if it only granted increased Attack Speed.

The Diamond Arrow is what makes that matter: all its Hits are Critical Hits, with 60% increased Critical Damage Bonus. On a bow character sitting at 28.88% Critical Hit Chance, that Arrow takes Total DPS from 120,714 to 297,952.

The six Arrows are rolled per Arrow and behave too differently to average into one number, so the item's line only sets a flag and a new config picks which Arrow to calculate. This follows the existing "Random element mode" config for modifiers that choose an element. The default, "Random (none applied)", applies none of them, so no existing build changes.

Each Arrow reuses modifiers that already exist:

| Arrow | Modifiers |
| --- | --- |
| Crescendo | ChainCountMax +6, 60% increased Damage per Chain |
| Splinter | SplitCount 6 |
| Reversing | PierceAllTargets |
| Diamond | CritChance 100 override, 60% increased CritMultiplier |
| Covetous | 600% increased LootRarity |
| Blunt | 600% increased EnemyHeavyStunBuildup |

The Reversing Arrow also returns to you, which PoB does not model.

- Equipped Cadiro's Gambit on a level 96 Ice Shot Deadeye and stepped through every option of the new config.
- Diamond: Total DPS 120,714 -> 297,952, Crit Chance 28.88% -> 100.00%, Crit Multiplier 396% -> 456%. Screenshots below.
- Crescendo: Chain Max 0 -> 6, and with "# of times Skill has Chained" set to 6, Full DPS 392,738 -> 651,538.
- Splinter: Split Count 0 -> 6. Reversing: Pierce Count 0 -> "All targets". Neither changes single target DPS, as expected.
- Regenerated ModCache.lua, which now caches the line as a parsed flag instead of an unsupported mod.

Any bow build with this item works. The quiver used above:

```
Rarity: UNIQUE
Cadiro's Gambit
Primed Quiver
Item Level: 82
LevelReq: 66
Implicits: 0
10% increased Attack Speed
Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow
```

### Before screenshot:
Path_of_Building-PoE2_Mv7eOzs2zW

### After screenshot:
Path_of_Building-PoE2_pn4cFVBIM8

### DIFF
--- a/src/Data/ModCache.lua
+++ b/src/Data/ModCache.lua
@@ -5226,7 +5226,7 @@ c["Double Stun Threshold while Shield is Raised"]={{[1]={[1]={globalLimit=100,gl
 c["Double the number of your Poisons that targets can be affected by at the same time"]={{[1]={flags=0,keywordFlags=0,name="PoisonCanStack",type="FLAG",value=true},[2]={flags=0,keywordFlags=0,name="PoisonStacks",type="MORE",value=100}},nil}
 c["Drop Ignited Ground while moving, which lasts 8 seconds and Ignites as though dealing Fire Damage equal to 10% of your maximum Life"]={nil,"Drop Ignited Ground while moving, which lasts 8 seconds and Ignites as though dealing Fire Damage equal to 10% of your maximum Life "}
 c["Drop Shocked Ground while moving, lasting 8 seconds"]={nil,"Drop Shocked Ground while moving, lasting 8 seconds "}
-c["Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow"]={nil,"Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow "}
+c["Each Arrow fired is a Crescendo, Splinter, Reversing, Diamond, Covetous, or Blunt Arrow"]={{[1]={flags=0,keywordFlags=0,name="Condition:PerandusArrows",type="FLAG",value=true}},nil}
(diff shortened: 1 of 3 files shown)


lucaskarsten:
Pushed a fix: the config now gates itself on `Condition:PerandusArrows`, granted by the quiver.

`ifFlag` only decides whether the control is drawn (`ConfigTab.lua:445`), while `ConfigTab:BuildModList` applies every stored `list` var regardless (`ConfigTab.lua:902`). Without the condition, a build that picked Diamond and then unequipped the quiver kept `CritChance OVERRIDE 100` with the control no longer visible — measured 100% Critical Hit Chance on a character with no Cadiro's Gambit equipped. Every other item-gated `ifFlag` option in the file only sets a `Condition:` or `Multiplier:` that is inert without its item, so this was the odd one out.

With the quiver equipped the numbers are unchanged (Diamond still 903,906 Full DPS on the character in the description); without it, the character measures its normal 29.45% Critical Hit Chance again.

`ModCache.lua` regenerated with `REGENERATE_MOD_CACHE=1 luajit HeadlessWrapper.lua`. Full busted suite: 566 successes / 0 failures / 0 errors.

PR #2419, opened 2026-08-15
Fix Exposure mod on Whisper Ice not working

### LINKED ISSUE #2379: Elemental Exposure on Hit mod from The Whispering Ice is not applied
### Check version

### Check for duplicates

### Check for support

Windows

The game shows elemental exposure on the monsters when I hit them, exactly the amount they should have. My staff applies -60% elemental exposure on hit. Which shows as -30% on unique bosses.

Calcs screen does not show elemental exposure cause by The Whispering Ice.

1. Start a new build.
2. Add unique item The Whispering Ice, Permafrost Staff.
3. In Skills, add Elemental Weakness (just to have a resist debuff showing in Calcs).
4. In Calcs, select Icestorm -> Storm.
5. Move the mouse over Elemental Weakness in Curses and Debuffs. You will see only the resist debuff applied by EW and no additional resist mods.
6. In Skills, add Frost Bomb.
7. Go back to Calcs and see the exposure from Frost Bomb showing correctly. But no exposure from The Whispering Ice.

The mod that applies exposure on hit on The Whispering Ice is not marked as "Not Supported in PoB". Therefore, there must be a bug that prevents it from getting included in Calcs.

### PoB for PoE2 build code

```shell
Example of a real build affected by this issue:
https://pobb.in/Igv-YtdqJ4qk
```

### Screenshots

_No response_

lucaskarsten:
Fixes #2379.

The reported symptom is that The Whispering Ice's exposure does not show up in the Calcs tab. The exposure is in fact applied — with only a `Permafrost Staff` carrying the line equipped, the enemy ends at Fire Resistance 20 on the default Pinnacle Boss config: 50 from the enemy config plus `-30` sourced to the item, the 60% halved against a unique enemy, which matches the game.

What is wrong is the state PoB is left in. Measured on `dev` with that item alone:

```
enemy Condition:HasFireExposure   true
player Condition:CanApplyFireExposure   nil
```

`hasExposureSource` (`Modules/CalcPerform.lua:429`) only looks at the player side: an `<Element>ExposureChance` modifier or the `InflictExposure` flag. The Whispering Ice grants exposure straight to the enemy through `EnemyModifier`, and so does `Enemies in your Presence have Exposure`; neither carries a chance modifier or the flag, so no source is found.

`Condition:CanApply<Element>Exposure` is what gates the `Is the enemy Exposed to Fire/Cold/Lightning?` config options (`Modules/ConfigOptions.lua:1866-1872`), so with those items the options are never drawn and there is nowhere in the Calcs tab to see the exposure, even though the resistances already reflect it.

The enemy's own exposure modifiers now count as a source.

- Added `spec/System/TestExposureSource_spec.lua`: the item grants a source for all three elements, a build with no exposure still reports none, and enabling the config on top of the item does not change the enemy's resistance. Against `dev` it reports 2 successes / 1 failure; with the fix, 3 successes / 0 failures.
- Exposure resolves to the largest single source (`CalcPerform.lua:3406`), so the config's -20% cannot stack onto the item's -60%; the third test covers that directly.
- Full busted suite: 569 successes / 0 failures / 0 errors.


LocalIdentity (approved):
The issue was in ModParser, not calc perform
This is a vastly simpler fix

PR #2432, opened 2026-08-25
Add Path of Exile 2 BuildPlanner (.build) export

### LINKED ISSUE #1829: Feature suggestion: Export to GGG's .build (in-game Build Planner) format
Hi PoB team, thanks for keeping the PoE2 fork active. Wanted to float something that came up while I was building adjacent tooling.

GGG ships a new .build file format consumed by PoE2's in-game Build Planner. The format is documented in their Developer Docs → Build Planner section (https://www.pathofexile.com/developer/docs/game). Currently PoB doesn't export to it, so users who want their PoB-designed build to surface in-game have to round-trip through external tooling.

Why it might be worth considering

Native export from PoB would make the workflow seamless: design in PoB → click Export → drop file in Documents\My Games\Path of Exile 2\BuildPlanner\ → load in-game

Prior work (not asking you to use it, just noting it exists)

I built a third-party converter: https://github.com/chesler410/poe2-build-forge (live at https://chesler410.github.io/poe2-build-forge/). It takes a PoB XML export and outputs a schema-validated .build file.

The non-trivial part is mapping PoB's integer tree node IDs to GGG's PassiveSkills.id strings — I use repoe-fork/poe2's Default.json (https://github.com/repoe-fork/poe2/blob/master/data/passive_skill_trees/Default.json) for that lookup. PoB has its own GGPK-extracted data that would serve the same purpose internally. The schema itself is captured at packages/schema/src/poe2-build.schema.json (https://github.com/chesler410/poe2-build-forge/blob/main/packages/schema/src/poe2-build.schema.json) if useful as a reference.

Asking

1. Is .build export on the roadmap already in any form? Don't want to duplicate effort.
2. If it'd be welcome as a community contribution, happy to describe the mapping algorithm in more detail and/or help test against PoB's existing data. (My code is TypeScript; PoB is Lua so a direct port doesn't make sense — but the algorithm is the small part.)

Either way, no pressure. Feel free to close out if it's out of scope.

Thanks!

vaisest:
Fixes #1829. This is an extension of and an update to #1832 by @TheClassified and @Wires77.

See #1832. This makes many changes to the previous PR:

- Add markup buttons to note menu
- Fix thorns crashing due to missing gem data
- Avoid notes overwriting item text
- Add colour to item text (quality, magic mod colour, fractured etc)
- Remove a lot of dead code
- Convert exporting to export single loadouts.
  This happens either by selecting specific sets,
  or by exporting all loadouts. The GGG file format
  is not really compatible with putting a whole leveling
  experience in one file due to level interval limitations.
  This now matches what build sites are currently doing more closely.

The changes mean that this feature does NOT include any level details in the build file. The previous approach of putting multiple loadouts in one file and filtering them using the level intervals does not work. We could add per-node/item/gem level filters, but I don't really see how they are helpful. As far as I know they only hide things in-game, which isn't really helpful. It is imo a better idea to just point out level pathing in notes if it is important.

The `unique_name` field was removed as it seemed useless. If this is shown in a special way somewhere on the UI, it can be added back.

Opinions welcome. I am not a build guide author. Code opinions are also welcome. This PR has gone through a lot of iteration and some parts of it may not be in line with others.

- Some example builds seem to work.
- Probably needs a lot more testing

Todo:

- avoid opening gem selector when opening note menu

https://poe.ninja/poe2/pob/278b8

### After screenshot:

image
image
image
image
image
image


LocalIdentity (approved):

LocalIdentity (approved):
Looks good now

PR #2445, opened 2026-08-28
Add support for select pseudo stats in trader weights

vaisest:

https://github.com/PathOfBuildingCommunity/PathOfBuilding/pull/10085

- Automated tests
- Manual testing

### After screenshot:

image


vaisest:
~~Closed due to a fundamental issue with augments on the trade site. My intention with the rune options has been to exclude them from the evaluation completely, and this ruins that because they would be included in the weighted sums. People do a lot of stat padding with res runes so this would not be good at all~~

image

Actually, adding this for parity reasons, but setting the map to be empty. This way it's easier to change direction later and porting pob1 changes stays easy

LocalIdentity (approved):

PR #2471, opened 2026-08-29
Tooltip support for keywords

cupkax:
This is a two fold PR which adds support for descriptive text to keywords on the passive tree. 

There are two (informal) tiers I have found which this useful for. Several ascendancy nodes and annointable passives have text saing `Grants X` e.g. `Grants Unravelling` or `Grants Inevitable Critical Hits`. Now, in-game this is properly parsed as we just click on the keyword, however, in pob there is no extra info as to what these nodes do. This PR now properly shows this text on hover, essentially piping the in-game as part of the tooltip.

Second issues is regarding the plethora keyword support which poe2 added ranging from commonly stuff like `Armour` to `Volatility`. As these keywords are part of larger context they are often found alongside each other. THis PR now provides a desciption for each of the keywords found in the tooltip (currently mapped to ALT). The tooltip itself now underlines the keywords for clarity. And as this can get annoying, I have added a toggle in the options panel to disable the keyword-specific tooltips if needed.

The current implementation creates a side-panel for each of the keywords and lists them vertically. The biggest culprit where this can be overwhelming is when a keyword has a very large description, `stun`, and/or when several keywords exists alongside each other `One with the River` node.

- Manually checked in PoB

### After screenshot:

image1

image


vaisest:
I feel like it would make sense to use the GGG keyword popup texts (`KeywordPopups.dat`) for this? That'd automatically include any changes GGG makes

cupkax:
I am guessing we need an exporter for this to get the strings? I haven't dabbled into that side of things.

cupkax:
It'd actually be great to have nested, pinnable tooltips like we have in game. Not sure if that's something which SG handles though

vaisest:
I meant just taking the description text. Popups would be nice but probably very difficult to implement.

I can get you the export later today. I had it in a branch somewhere.

phulsandhu13-pixel (commented):
##- Please type your reply above this line -##
I’m not sure if it’s a coincidence or a fact that

cupkax:
@LocalIdentity Could you take a look at this when you are free? I've tried to curate the deny list of keywords to reduce annoyance for end-users and is currently implemented as a side-panel so the damage calculations are not croweded. There is also an options to toggle off the key-based descriptions as needed

PR #2497, opened 2026-09-01
Added support for Decree of Loyalty mods

ntcarlson:
Add support for the following Decree of Loyalty unique modifiers:
* Convert 1% of maximum Life to twice as much Armour per 1% Chaos Resistance above 0%
* Defend with (150-200)% of Armour while you have Energy Shield

I left the final "Damage over time cannot bypass Energy Shield" modifier unimplemented because I am not  sure how to handle it.

- Verified that life and armour values match in-game values on several test characters, including one with more than 100% life converted to other stats.

### Before screenshot:
20260831_19h20m53s_grim

### After screenshot:
20260831_19h34m48s_grim
20260831_19h34m19s_grim
20260831_19h34m09s_grim

### DIFF
--- a/src/Modules/ModParser.lua
+++ b/src/Modules/ModParser.lua
@@ -6311,10 +6311,10 @@ local specialModList = {
 	} end,
 	["you can socket an additional copy of each lineage support gem, in different skills"] = { mod("MaxLineageCount", "BASE", 1) },
 	["you can socket (%d+) additional copies of each lineage support gem, in different skills"] = function(num) return { mod("MaxLineageCount", "BASE", num) } end,
-    ["convert 1%% of maximum life to twice as much armour per 1%% chaos resistance above 0%%"] = { 
-		mod("LifeConvertToArmour", "BASE", 1, { type = "PerStat", stat = "ChaosResist"}),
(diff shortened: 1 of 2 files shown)


src/Modules/ModParser.lua
```diff
@@ -6308,6 +6311,10 @@ local specialModList = {
 	} end,
 	["you can socket an additional copy of each lineage support gem, in different skills"] = { mod("MaxLineageCount", "BASE", 1) },
 	["you can socket (%d+) additional copies of each lineage support gem, in different skills"] = function(num) return { mod("MaxLineageCount", "BASE", num) } end,
+    ["convert 1%% of maximum life to twice as much armour per 1%% chaos resistance above 0%%"] = { 
+		mod("LifeConvertToArmour", "BASE", 1, { type = "PerStat", stat = "ChaosResist"}),
```
LocalIdentity: This should use a function to grab the convert and the per chaos res value, not use hard coded 1% for each

LocalIdentity (changes requested):

src/Modules/ModParser.lua
ntcarlson: Alright, I corrected that issue. I wasn't sure whether the numbers should be integers or decimals, but I made it accept decimal values as well

LocalIdentity (approved):

PR #2499, opened 2026-09-02
Fix a bunch of unique export source files

vaisest:

Adds the other armour type inc lines to the item db unique. The second commit also fixes a bunch of mod ids by cross-referencing with Rasmus' json. This does update values for some items, too.

There are a lot more incorrect ones (200+)  which have the same values, but I didn't want to go through them here

- Item manually tested

### After screenshot:
image

### DIFF
--- a/src/Data/Uniques/shield.lua
+++ b/src/Data/Uniques/shield.lua
@@ -223,7 +223,7 @@ Spiked Buckler
 League: Dawn of the Hunt
 Implicits: 1
 Grants Skill: Parry
-(50-80)% increased Evasion Rating
+(60-100)% increased Evasion Rating
 +(10-15)% to all Elemental Resistances
 Parrying applies 10 Stacks of Critical Weakness
 100% increased Parry Damage
@@ -353,7 +353,7 @@ Variant: Current
 Implicits: 1
 Grants Skill: Raise Shield
 {variant:1,2}(30-40)% increased Block chance
-{variant:3}(10-15)% increased Block chance
+{variant:3}(15-20)% increased Block chance
 (30-50)% increased Armour and Energy Shield
 {variant:1}+(15-30) to maximum Mana
 {variant:2,3}+(50-70) to maximum Mana
--- a/src/Data/Uniques/shield.lua
+++ b/src/Data/Uniques/shield.lua
@@ -137,12 +137,15 @@ for 4 seconds, every 0.25 seconds while raised
 ]],[[
 Wulfsbane
 Painted Tower Shield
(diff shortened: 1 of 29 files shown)


vaisest:
e9642b6 includes a correction for mod IDs which differed from the Ninja json, but had the same mod text. This was done by a clanker, but I did a pass after it, ensuring that the only IDs not found in the json are legacy mods. The first pass did NOT change any mod text, but in the latter I found some issues, some of which did change the generated mod text.


40d7049 fixes some mods which were incorrect in PoB. Some of these seemed to be missing from the wiki, too.

The used script:

```python
import pathlib
import lupa
import json
import re

lua = lupa.LuaRuntime(unpack_returned_tuples=True)
exportFolder = pathlib.Path("../src/Export/Uniques/")

uqJson = json.loads(pathlib.Path("../uniques.json").read_text())

# dunno why these are missing
ignored = ["Split Personality", "Tabula Rasa", "Guiding Palm", "Palm of the Dreamer"]


def getEntry(name):
    return next(entry for entry in uqJson["uniques"] if entry["name"] == name)


for file in exportFolder.iterdir():
    contents = file.read_text()
    table = lua.eval(f"(function()\n{contents}\nend)()")
    for item in table.values():
        title = item.split("\n")[0]
        if title in ignored or "Source: No longer obtainable" in item:
            continue
        entry = getEntry(title)
        entryIds = [mod["modId"] for mod in entry["mods"]]
        for line in item.split("\n")[2:]:
            line = re.sub(r"\{.+\}", "", line)
            if (
                line != ""
                and ":" not in line
                and "[" not in line
                and line not in entryIds
            ):
                print(title, line)
```

src/Data/Uniques/shield.lua
```diff
@@ -137,12 +137,15 @@ for 4 seconds, every 0.25 seconds while raised
...
+Version: Pre 0.3.0
+Version: Current
 Implicits: 1
 Grants Skill: Raise Shield
 +(40-60) to maximum Life
 +(10-15) to Strength
 +(60-80) to Stun Threshold
-Permanently Intimidate enemies on Block
+{version:2}Intimidate Enemies on Block for 8 seconds
+{version:1}UniqueEnemiesBlockedAreIntimidated
```
LocalIdentity: Need to fix this export

LocalIdentity (changes requested):

src/Data/Uniques/shield.lua
vaisest: Oops, missed the "1". Fixed in 3e9aa40

PR #2503, opened 2026-09-06
Fix LuaJIT compatibility on dev

JLawMcGraw:

Recent dev changes introduced syntax accepted by Luau but not by the LuaJIT 2.1 runtime used by Path of Building and its test workflow. Attribute-node switching also recursively deep-copies references into the passive-tree graph, which greatly increases allocation and can exhaust LuaJIT table capacity during replay-heavy workloads.

This replaces the incompatible operators, optional chaining, continue statements, and arrow function with LuaJIT-compatible equivalents. It also switches attribute nodes with a shallow structural copy while copying the stat-description array that ReplaceNode mutates.

- Parsed all 1,121 Lua files with stock LuaJIT 2.1.
- Ran all 53 system spec files: 824 passed, 0 failed, 2 expected failures, 18 pending.
- Swept all 293 attribute nodes and verified each override while reducing allocation from about 5 MB to about 993 KB.

N/A; this is runtime compatibility and passive-tree correctness.

### Before screenshot:

N/A

### After screenshot:

N/A

### DIFF
--- a/spec/System/TestItemMods_spec.lua
+++ b/spec/System/TestItemMods_spec.lua
@@ -202,13 +202,13 @@ describe("TetsItemMods", function()
 		build.itemsTab.controls.craftingSorting:SelByValue("Life", "stat")
 		build.calcsTab.GetMiscCalculator = function()
 			return function(args)
-				calcCount += 1
+				calcCount = calcCount + 1
 				local life = 0
 				for _, modLine in ipairs(args.repItem.explicitModLines) do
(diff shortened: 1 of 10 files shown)


vaisest:
This syntax is part of LuaJIT v2.1 as of July 2026. Update your LuaJIT: https://github.com/LuaJIT/LuaJIT/issues/1476

LocalIdentity:
We use a newer version of LuaJIT so the syntax changes aren't needed. Open a new PR with the attribute node memory changes

PR #2505, opened 2026-09-06
Export 0.5.5 data and update trade data

vaisest:

Fixes export schema according to https://github.com/poe-tool-dev/dat-schema/commits/main/dat-schema/poe2/_Core.gql

It also seems that there were some GGG-side asset changes. These were fixed via some LLM consultation as I don't know much about how the DDS things work. The fixes are f953211 and 1296733, and ended up being fairly simple.

bef4e70 fixes a var name that very much looks like a latent bug that caught my eye randomly

- Tests fixed and now pass (I verified the stat scaling in-game. 120% seemed like an odd case, but it does check out)
- Assets fixed and work properly


vaisest:
b3282b7 also makes a change to the LiquidEmotionOutcomes column order in spec.lua. These seem to have been swapped in the 0.5 export, which meant that sapphire mods were on emerald jewels and vice versa.

LocalIdentity (approved):

PR #2537, opened 2026-09-19
Fix unintended utf8.reverse argument

meehl:
`utf8.gsub` returns both the substituted string and the number of substitutions. Since the gsub calls are passed directly to `utf8.reverse`, both return values get forwarded, so the substitution count ends up as the second argument which is then interpreted as the optional `lax` argument.

This wraps the gsub calls in parentheses so only the substituted string is passed to `utf8.reverse`.


meehl:
This fix also needs to be applied to PoB1.

PR #2540, opened 2026-09-20
Fix Runeforged/Runemastered base variants never being added to uniques

515k4:
Regression from #2511 — no issue filed.

The Runeforged/Runemastered base variants added in #2511 never appeared on any unique, so the whole feature was inert.

The unique DB loader in `src/Modules/Main.lua` skips injecting the runeforging bases when the unique already declares base variants of its own, but it tested that with `newItem.baseLines`:

```lua
local hasBaseVariants = newItem.baseLines and not not next(newItem.baseLines)
```

`ItemClass:ParseRaw` fills `baseLines` with the item's *own* base on every parse (`src/Classes/Item.lua`, `self.baseLines[baseName] = { ... }`), and the enclosing `if newItem.base then` already guarantees a base was recognised. So `hasBaseVariants` was true for every unique and the injection below it was dead code.

`newItem.baseList` is the field that actually answers the question: it stays `nil` unless a `Base Variant:` line is present. The guard now tests that instead. The `newItem.baseList = newItem.baseList ?? {}` a few lines down was both mis-indented and now provably always nil-coalescing, so it became a plain `= { }`.

Measured headless, before the fix: **0 of 443** uniques came back with a `baseList`, while **272** were eligible (their base has a `Runeforged `/`Runemastered ` counterpart in `data.itemBases`). After the fix all 272 get one.

The existing tests missed this because `TestItemVariants_spec.lua` feeds in raw text that already declares `Base Variant:` lines — it exercises the item machinery, which works, and never the loader. The added tests go through `main.uniqueDB`.

- Added three tests in `spec/System/TestItemVariants_spec.lua` that read from `main.uniqueDB` rather than raw text: Wanderlust gets `{ "Regular Base", "Runeforged", "Runemastered" }` with correctly tagged `baseLines` and `selectedBase == 1`; the variants survive a raw round trip and selecting base 3 resolves to `Runemastered Wrapped Sandals`; and **every** eligible unique gets a `baseList`. All three fail on `dev` and pass with the fix.
- Ran the full `spec/System` suite in `ghcr.io/pathofbuildingcommunity/pathofbuilding-tests:latest`: 51 specs, 0 failures, 0 errors.
- Regenerated ModCache with `REGENERATE_MOD_CACHE=1 luajit HeadlessWrapper.lua`; `git diff src/Data/ModCache.lua` is empty, so parsing the 272 extra base-variant items causes no cache churn.

N/A — reproduced headlessly; the numbers above come from iterating `main.uniqueDB.list`.

### Before screenshot:

image

### After screenshot:

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vaisest:
Hey, thanks for the report. You're correct that I intended for the variants to be automatic, but it seems this bug went unnoticed in the original PR, because adding them like this is actually incorrect. According to @LocalIdentity it should be done using the Verisium crafts dat table.

There's also another issue in the export scripts in that they're keyed by name, which means the uniques that have multiple runemastered base options aren't working in the program in general. I'll have a look at reworking this code in a PR soon

515k4:
> Hey, thanks for the report. You're correct that I intended for the variants to be automatic, but it seems this bug went unnoticed in the original PR, because adding them like this is actually incorrect. According to @LocalIdentity it should be done using the Verisium crafts dat table.
> 
> There's also another issue in the export scripts in that they're keyed by name, which means the uniques that have multiple runemastered base options aren't working in the program in general. I'll have a look at reworking this code in a PR soon

Thank for pointing this out it. In that case I will close this PR and wait for your proper fix. Btw do you need any help on it?

